PR #7, opened 2025-09-10
Feature/add kover basics with initial threshold

eslam-99:
Added kover basic setup following module-based by applying kover configuration and threshold for every testable module

- Added kover for each testable module without exceptions for now
- Set threshold by 80% for each module

### DIFF
--- a/trends_presentation/build.gradle.kts
+++ b/trends_presentation/build.gradle.kts
@@ -5,6 +5,7 @@ plugins {
     alias(libs.plugins.composeMultiplatform)
     alias(libs.plugins.composeCompiler)
     alias(libs.plugins.androidLibrary)
+    alias(libs.plugins.kover)
 }
 
 kotlin {
@@ -42,6 +43,9 @@ kotlin {
         iosMain.dependencies {
 
         }
+        commonTest.dependencies {
+            implementation(libs.kotlin.test)
+        }
     }
 }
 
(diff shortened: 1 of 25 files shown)


AhmadKharfan (approved):

AimanYosofiUlt (approved):

ahmednasserzaza (approved):
Well done

trends_presentation/build.gradle.kts
```diff
@@ -52,4 +56,12 @@ android {
     defaultConfig {
         minSdk = libs.versions.android.minSdk.get().toInt()
     }
+}
+
+kover.reports {
+    verify {
+        rule {
+            minBound(80)
```
mohamedd-hassan, line 64: I understand Kover reporting and unit testing in presentation for viewmodel, but wouldn't 80% here be too high as composables will make up the majority of the module leading to the threshold not being reached.

or is this just an initial setup so that each team can then set the exclusion to the composable packages independently?

AhmadKharfan: This is just an initial threshold for now we can adjust it later based on our discussions and agreement on what makes sense for each module

eslam-99: As Ahmed mentioned, it's just initial setup to avoid blocking other CI/CD tasks

ThoriaHamdy99 (approved):

PR #8, opened 2025-09-10
Feature/add domain and data abstraction

ahmedmahgoub0:
### Description
- add `Interest` entity
- add `InterestsRepository` abstraction

### DIFF
--- a/trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/entity/Interest.kt
+++ b/trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/entity/Interest.kt
@@ -1,0 +1,7 @@
+package net.thechance.mena.trends.domain.entity
+
+data class Interest(
+    val id: Int,
+    val name: String,
+    val emoji: String
+)
--- a/trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/repository/InterestsRepository.kt
+++ b/trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/repository/InterestsRepository.kt
@@ -1,0 +1,9 @@
+package net.thechance.mena.trends.domain.repository
+
+import net.thechance.mena.trends.domain.entity.Interest
+
+interface InterestsRepository {
+    suspend fun getAllInterests(): List<Interest>
+    suspend fun checkUserHasInterests(): Boolean
+    suspend fun updateUserInterests(interestsIds: List<Int>)
+}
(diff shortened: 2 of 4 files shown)


iZyadNasser (approved):

ahmednasserzaza (approved):

ahmeddsamii (approved):

mohamedd-hassan (approved):

Shahd-Hatem (approved):

fatimafaisalghazi (approved):

trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/repository/InterestsRepository.kt
```diff
@@ -0,0 +1,9 @@
+package net.thechance.mena.trends.domain.repository
+
+import net.thechance.mena.trends.domain.entity.Interest
+
+interface InterestsRepository {
+    suspend fun getAllInterests(): List<Interest>
+    suspend fun checkUserHasInterests(): Boolean
```
skaik-mo: I think we should fetch the user’s interests, check if they are empty, and then present the first screen `First Time User(Onboarding for Trends)` when no interests are found.

ahmedmahgoub0: but should I get the user interests list each time just to know if user has interests?

MahmoodTarek (approved):

trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/repository/InterestsRepository.kt
iZyadNasser: @skaik-mo There are two points here:
1. I don't think we should fetch all user interests and load them via network just to use them for a simple true/false query, this will be an unnecessary load on the network.

2. Even if, for the sake of argument, we used that approach (we fetched all user data and checked if they are empty), we can do this inside the repository itself, then in the end return a Boolean of true/false to the ui, the ui does not have to get all the data and do the logic by itself, all it needs is the Boolean value, so we can abstract that for the presentation layer and do the logic inside the repository. Again, all that is if we fetch all user interests, which I don't think is correct in any case.

So, in both approaches, the repository should in the end only provide a Boolean from the function, its up to the repository implementation and the backend implementation on how this Boolean will be created.

trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/entity/Interest.kt
```diff
@@ -0,0 +1,7 @@
+package net.thechance.mena.trends.domain.entity
+
+data class Interest(
```
ahmeddsamii: i think this should be category because interest isn't understandable

HendSayed25 (approved):

trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/entity/Interest.kt
ahmedmahgoub0: agreed with you

PR #9, opened 2025-09-10
Feature/create base view model for chat

Joseph-Sameh-0:


core_chat_presentation/src/commonMain/kotlin/net/thechance/mena/core_chat/presentation/shared/BaseViewModel.kt
```diff
@@ -0,0 +1,74 @@
...
+        }
+        return coroutineScope.launch(exceptionHandler + dispatcher) {
+            onStart()
+            try {
+                val result = execute()
+                onSuccess?.invoke(result)
+            } catch (e: Exception) {
+                onError(e)
+            }
+        }
```
ZainabAlnajjar: I believe that `exceptionHandler` is enough to handle any exception, is there a reason for adding another `try catch` block?

dukan_domain/build.gradle.kts
```diff
@@ -3,6 +3,7 @@ plugins {
 }
 
 kotlin {
+    jvm()
```
ahmednasserzaza: jvm has been added to 'develop' branch , just pull the latest update

core_chat_presentation/src/commonMain/kotlin/net/thechance/mena/core_chat/presentation/shared/BaseViewModel.kt
Joseph-Sameh-0: no the CoroutineExceptionHandler will not catches all exceptions it just catches the coroutine exceptions

core_chat_presentation/src/commonMain/kotlin/net/thechance/mena/core_chat/presentation/shared/BaseViewModel.kt
```diff
@@ -0,0 +1,74 @@
...
+            onError(throwable)
+        }
+
+        return coroutineScope.launch(exceptionHandler + dispatcher) {
+            try {
+                collect()
+                    .onStart { onStart() }
+                    .catch { onError(it) }
+                    .onEmpty { onEachEmit(null) }
+                    .collectLatest { onEachEmit(it) }
```
ZainabAlnajjar: I think `onCollect `could be better naming instead of `onEachEmit`, also why do we need the `onEmpty`? 
I think it could be better to just pass `onCollect` with not-nullable data type, and handle it in the viewmodel instead of handling null

ahmednasserzaza (approved):
Well done,
It would be better if we add a description for our PRs, even if it has a small changes.

dukan_domain/build.gradle.kts
Joseph-Sameh-0: the base branch is not the develop

monaayman123 (approved):

core_chat_presentation/src/commonMain/kotlin/net/thechance/mena/core_chat/presentation/shared/BaseViewModel.kt
Joseph-Sameh-0: the collectLatest don't work with empty flow and will not call onEachEmit so the null here will indicate that the flow was empty

ZainabAlnajjar (approved):

PR #14, opened 2025-09-11
Feature/[TRN-8] add Trend entity and TrendRepository interface

MahmoodTarek:
## Description:
- Added `Trend` entity.
- Added `TrendRepository` interface.
- Added `FakeReelsRepositoryImpl`


trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/entity/Trend.kt
```diff
@@ -0,0 +1,11 @@
...
+
+data class Trend(
+    val id: Int,
+    val thumbnailUrl: String,
+    val videoUrl: String,
+    val likesCount: Int,
+    val viewsCount: Int,
+    val description: String,
+    val createdAt: Long
+)
```
ahmeddsamii: This should have category to know the categories of trend

trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/entity/Trend.kt
```diff
@@ -0,0 +1,11 @@
+package net.thechance.mena.trends.domain.entity
+
+data class Trend(
+    val id: Int,
+    val thumbnailUrl: String,
+    val videoUrl: String,
+    val likesCount: Int,
+    val viewsCount: Int,
+    val description: String,
+    val createdAt: Long
```
ahmedmahgoub0: Can we use Kotlin local date instead of long to represent the created date?

trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/entity/Trend.kt
```diff
@@ -0,0 +1,12 @@
+package net.thechance.mena.trends.domain.entity
+
+data class Trend(
+    val id: Int,
+    val thumbnailUrl: String,
+    val videoUrl: String,
+    val description: String,
+    val likesCount: Int,
+    val viewsCount: Int,
+    val createdAt: Long,
```
ahmednasserzaza: It should be better if we made it a LocalDateTime from kotlin istead Long

ahmednasserzaza (changes requested):

ahmeddsamii (approved):

ahmedmahgoub0 (approved):

HendSayed25 (approved):

mohamedd-hassan (approved):

trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/entity/Trend.kt
iZyadNasser: @ahmeddsamii I think we could make it so that there is a function to getCategoriesByTrendId(), instead of including the categories list inside the entity model, we should try to keep the entity as simple as possible I think

trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/entity/Trend.kt
```diff
@@ -0,0 +1,14 @@
+package net.thechance.mena.trends.domain.entity
+
+import kotlinx.datetime.LocalDateTime
+
+data class Trend(
```
iZyadNasser: Is "Trend" the best naming for the entity? I personally did not understand what you mean by this entity until I read the properties and understood that it represents our short videos

iZyadNasser: Also, do we need all the properties you added to this model in the current sprint? I believe we can remove some of them for now and only add them when we need them, or what do you think?

ahmednasserzaza: @iZyadNasser totally agree with u

ahmednasserzaza (approved):

PR #16, opened 2025-09-11
Feature/Configure Automated Deployment with Firebase

ABDULLAHHG:
### Configure Automated Deployment with Firebase


.github/workflows/deploy_release.yml
AhmadKharfan, line 1: I think it will be better if we create a separate workflow just for building the release APK then here in the main workflow we can have jobs for different deployments Firebase, Slack, etc... this way if we want to deploy somewhere else we just add a new job without touching the build steps

.github/workflows/deploy_release_to_firebase.yml
```diff
@@ -0,0 +1,80 @@
...
+        env:
+          BUILD_TOOLS_VERSION: "35.0.0"
+
+      - name: Upload signed APK to Firebase
+        uses: wzieba/Firebase-Distribution-Github-Action@v1
+        with:
+          appId: ${{ secrets.FIREBASE_APP_ID }}
+          serviceCredentialsFileContent: ${{ secrets.CREDENTIAL_FILE_CONTENT }}
+          file: ${{ steps.sign_app.outputs.signedReleaseFile }}
+          debug: true
```
AhmadKharfan: You could create tester groups in Firebase and then set the groups parameter here to control which group of testers receives the app

AhmadKharfan (approved):
Well done Abdullah

PR #19, opened 2025-09-12
CHT-2-20 Device Contacts Datasource

dev-bilal-azzam:
- Add `contacts-provider` library for fetching Contacts from device.
- Implement `DeviceContactsDatasource` using `ContactsProvider`.
- Use `DeviceContactsDatasource` in `ContactsRepositoryImpl`.

### DIFF
--- a/core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/contacts/ContactMapper.kt
+++ b/core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/contacts/ContactMapper.kt
@@ -6,6 +6,7 @@ import net.thechance.mena.core_chat.data.shared.dto.PagedDataDto
 import net.thechance.mena.core_chat.domain.entity.Contact
 import net.thechance.mena.core_chat.domain.exception.ContactsFetchFailedException
 import net.thechance.mena.core_chat.domain.model.PagedData
+import com.bilalazzam.contacts_provider.Contact as DeviceContact
 
 fun PagedDataDto<ContactDto>?.toPagedListOfContacts(): PagedData<Contact> {
     val pagedData = this ?: throw ContactsFetchFailedException("Response body is null")
@@ -38,4 +39,18 @@ private fun Contact.toContactCreationRequestDto(): ContactCreationRequestDto {
         name = name,
         phone = phone
     )
+}
+
+fun List<DeviceContact>.toListOfContacts(): List<Contact> {
+    return flatMap { deviceContact ->
+        deviceContact.phoneNumbers.map { phone ->
+            Contact(
+                name = listOfNotNull(deviceContact.firstName, deviceContact.lastName)
+                    .joinToString(" "),
+                phone = phone,
+                isMenaUser = false,
+                imageUrl = null
+            )
+        }
+    }
 }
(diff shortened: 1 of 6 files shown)


core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImpl.kt
```diff
@@ -22,7 +24,8 @@ class ContactsRepositoryImpl(
         }.toPagedListOfContacts()
     }
 
-    override suspend fun syncContacts(contacts: List<Contact>) {
+    override suspend fun syncContacts() {
+        val contacts = deviceContactsDataSource.getDeviceContacts().toListOfContacts()
```
ZainabAlnajjar: I think this line could throw exceptions, for example if there is no permission, even if not now, the library could change by time 
It would be better to call it inside the try call

ZainabAlnajjar: Also I remember getting some contacts with just the name (no phone number) or just a number without a name, we need to filter them out, probably from the library itself, could you please confirm this case?

monaayman123 (approved):

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImpl.kt
dev-bilal-azzam: - The first case you are right it might throw exceptions.
- Second issue is already handled.

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/contacts/ContactMapper.kt
```diff
@@ -39,3 +40,17 @@ private fun Contact.toContactCreationRequestDto(): ContactCreationRequestDto {
...
+            Contact(
+                name = listOfNotNull(deviceContact.firstName, deviceContact.lastName)
+                    .joinToString(" "),
+                phone = phone,
+                isMenaUser = false,
+                imageUrl = null
+            )
+        }
+    }
+}
```
Joseph-Sameh-0: It's better to map to `List<ContactCreationRequestDto>` and use it directly in the repo without another mapping

Joseph-Sameh-0: Because in the repository we do `contacts.toListOfContactCreationRequestDto()`

dev-bilal-azzam: done

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImpl.kt
dev-bilal-azzam: done

Joseph-Sameh-0 (approved):

ZainabAlnajjar (approved):

Noor-Serry (approved):
good job

PR #28, opened 2025-09-13
refactor: create dukan base structure

YoussefmSaber:
## Changes in this PR:
- Replaced all magic numbers with enums.
- Handled on system back pressed.

### DIFF
--- a/dukan_presentation/src/iosMain/kotlin/net/thechance/mena/dukan/presentation/util/OnSystemBackPressed.ios.kt
+++ b/dukan_presentation/src/iosMain/kotlin/net/thechance/mena/dukan/presentation/util/OnSystemBackPressed.ios.kt
@@ -1,0 +1,7 @@
+package net.thechance.mena.dukan.presentation.util
+
+import androidx.compose.runtime.Composable
+
+@Composable
+actual fun OnSystemBackPressed(onBack: () -> Unit) {
+}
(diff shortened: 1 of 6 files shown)


dukan_presentation/src/iosMain/kotlin/net/thechance/mena/dukan/presentation/util/OnSystemBackPressed.ios.kt
```diff
@@ -0,0 +1,7 @@
+package net.thechance.mena.dukan.presentation.util
+
+import androidx.compose.runtime.Composable
+
+@Composable
+actual fun OnSystemBackPressed(onBack: () -> Unit) {
```
mostafa-n3ma, line 6: I don't see the actual iOS implementation ! ?

YoussefmSaber: the ios system it's self doesn't have a back button,  so there is no implementation for it

mostafa-n3ma: so why we are using the expect and actual in this case ?

YoussefmSaber: The reason for this, that we want to call this function on the devices that supports back handler, if we for example in the future implemented this for desktop with the same design as the mobile, the desktop doesn't have a back button, but the android devices has, so this is why we implemented this using expect and actual. I know it's kinda confusing, but it's really simple.

mostafa-n3ma (approved):

Mohamed-Ashraf-2210 (approved):

Ahmed-AlMancy (approved):

PR #29, opened 2025-09-13
Feature/Setup dependecy injection For trends

ahmeddsamii:
This Pr introduces Koin for dependency injection across the `trends_api`, `trends_data`, `trends_domain`, and `trends_presentation` modules.

Key changes include:
- Added Koin dependencies (`koin-core`, `koin-compose`, `koin-compose-viewmodel`, `koin-annotations`, `koin-ksp-compiler`) to the respective `build.gradle.kts` files.
- Configured KSP for Koin annotation processing in each module.
- Created Koin modules (`TrendApiModule`, `TrendDataModule`, `TrendDomainModule`, `TrendPresentationModule`) with `@Module` and `@ComponentScan` annotations to enable automatic dependency scanning.
- Annotated `TrendsApiImp` with `@Single(binds = [TrendsApi::class])` to register it as a singleton.
- Added an example `TestViewModel` demonstrating Koin ViewModel injection using `@KoinViewModel` and `@Provided`.
- Updated `TrendsNavGraph` to use a `TestScreen` as the initial destination.
- Made the `Route.Test` serializable.
- Removed the placeholder `.gitkeep` file from the `trends_presentation/screen` directory.
- Updated library versions for `navigationCompose`, `ksp`, `koin-core`, `koin-annotations`, `kotlinx-datetime`, and `kotlinx-serialization` in `gradle/libs.versions.toml`.


trends_data/build.gradle.kts
```diff
@@ -12,9 +23,41 @@ kotlin {
         }
         commonMain.dependencies {
             implementation(projects.trendsDomain)
+            implementation(libs.koin.core)
+            implementation(libs.koin.compose)
+            implementation(libs.koin.compose.viewmodel)
```
mohamedd-hassan: I wanted to ask, do we need the `koin.compose` and `koin.compose.viewmodel` here? as this is the data module so I don't think we need them

trends_domain/build.gradle.kts
```diff
@@ -9,7 +13,29 @@ kotlin {
 
     sourceSets {
         commonMain.dependencies {
-
+            implementation(libs.koin.core)
+            implementation(libs.koin.compose)
+            implementation(libs.koin.compose.viewmodel)
```
mohamedd-hassan: same here

trends_data/build.gradle.kts
ahmeddsamii: Yess, u are right i will fix this

ahmedmahgoub0, line 28: Could you please remove compose dependencies here for data module?

trends_api/src/commonMain/kotlin/net/thechance/mena/trends/api/TrendsApi.kt
```diff
@@ -0,0 +1,8 @@
+package net.thechance.mena.trends.api
+
+import androidx.compose.runtime.Composable
+
+interface TrendsApi {
+    @Composable
+    fun TrendsContainer()
```
ahmedmahgoub0: could we use another naming here for this function, thinking of starting with a verb that refers to what it does, like: `enterTrends`, `startTrends`, `launchTrends` ?

ahmeddsamii: In the meeting we agreed on this name i guess, but yeah i will change it

ahmednasserzaza: I agree with @ahmedmahgoub0, it would be more readable if we named it to 'launch', it would be totally enough

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/navigation/TrendsApiImp.kt
```diff
@@ -0,0 +1,19 @@
+package net.thechance.mena.trends.presentation.navigation
+
+import androidx.compose.runtime.Composable
+import androidx.navigation.compose.rememberNavController
+import net.thechance.mena.trends.api.TrendsApi
+import org.koin.core.annotation.Single
+
+@Single(binds = [TrendsApi::class])
+class TrendsApiImp () : TrendsApi {
```
ahmednasserzaza: rename it to TrendsApiImpl

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/navigation/TrendsNavGraph.kt
```diff
@@ -0,0 +1,22 @@
+package net.thechance.mena.trends.presentation.navigation
+
+import androidx.compose.runtime.Composable
+import androidx.navigation.NavHostController
+import androidx.navigation.compose.NavHost
+import androidx.navigation.compose.composable
+import net.thechance.mena.trends.presentation.screen.TestScreen
+
+@Composable
+fun TrendsNavGraph(
```
ahmednasserzaza: Rename it to TrendsNavHost because it's the parent and it may cosists of many navGraphs,
and just provide the navController here directly using CompositionLocalProvider and initiate navController as a parameter in top of the function, So it should be something like that 

```
fun TrendsNavHost() {

    val navController = rememberNavController()

    CompositionLocalProvider(
        LocalNavController provides navController
    ) {
        NavHost(
            modifier = Modifier.fillMaxSize(),
            navController = navController,
            startDestination = MainTrendsRoute,
        ) {
            composable<MainTrendsRoute> {
                MainTrendsScreen()
            }
        }
    }
}
```

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/base/BaseViewModel.kt
```diff
@@ -0,0 +1,59 @@
...
+import kotlinx.coroutines.flow.asSharedFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
+
+abstract class BaseViewModel<State, Effect>(
+    initialState: State
+) : ViewModel() {
+
+    private val _screenState = MutableStateFlow(initialState)
```
ahmednasserzaza: just state is enough

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/base/BaseViewModel.kt
```diff
@@ -0,0 +1,59 @@
...
+    val screenState: StateFlow<State> = _screenState.asStateFlow()
+
+    private val _effect = MutableSharedFlow<Effect>()
+    val effect = _effect.asSharedFlow()
```
ahmednasserzaza: search for throttleFirst in Rx java, then try to create an extension function that do the same behavior on Flow and use it here instead asSharedFlow directly

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/base/BaseViewModel.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    fun updateState(transform: State.() -> State) {
```
ahmednasserzaza: it should be protected

ahmednasserzaza: and updater is more readable than transform

ahmednasserzaza: You can also add a protected property that gets the current state

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/base/BaseViewModel.kt
```diff
@@ -0,0 +1,59 @@
...
+        _screenState.update { transform(it) }
+    }
+
+    protected fun sendEffect(
+        event: Effect,
+        onStart: suspend () -> Unit = {},
+        onEnd: suspend () -> Unit = {}
+    ) {
+        viewModelScope.launch(Dispatchers.IO) {
+            onStart()
```
ahmednasserzaza: onStart and onEnd is useless here, please remove them

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/base/BaseViewModel.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    protected fun <R> tryToCall(
+        block: suspend () -> R,
+        onSuccess: suspend (R) -> Unit,
+        onError: suspend (Throwable) -> Unit,
+        onStart: suspend () -> Unit = {},
+        onEnd: suspend () -> Unit = {},
+        dispatcher: CoroutineDispatcher = Dispatchers.IO
+    ): Job {
+        return viewModelScope.launch(dispatcher) {
```
ahmednasserzaza: use dispatcher + CoroutineExceptionHandler because it may throws an exception you didin't catch it, also we maybe excute the callee on a different courotineScope, So you can pass it as a parameter with a default value

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/base/BaseViewModel.kt
```diff
@@ -0,0 +1,59 @@
...
+        onEnd: suspend () -> Unit = {}
+    ) {
+        viewModelScope.launch(Dispatchers.IO) {
+            onStart()
+            _effect.emit(event)
+            onEnd()
+        }
+    }
+
+    protected fun <R> tryToCall(
```
ahmednasserzaza: tryToExcute is better 😅

ahmednasserzaza (changes requested):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/base/BaseViewModel.kt
ahmeddsamii: we are already catching the error by onFailure block so why do i need the CoroutineExceptionHandler here?

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/base/BaseViewModel.kt
ahmednasserzaza: sendEffect should only emit effects. Handling loading (onStart / onEnd) already exists in tryToExecute, so having them here duplicates responsibility and makes things confusing.

Suppose using loading state as an example

```
tryToExecute(
    block = { api.getData() },
    onStart = { updateState { copy(isLoading = true) } },
    onSuccess = { sendEffect(ShowSnackBar("Success")) },
    onEnd = { updateState { copy(isLoading = false) } }
)
```
Does we need an additional onStart, onEnd in sendEffect!

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/base/BaseViewModel.kt
ahmednasserzaza: As I remember, this didn’t cause any issues on Android when using try-catch. However, on iOS (and WASM), canceling the currently running scope leads to a crash.

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/util/extensions.kt
```diff
@@ -0,0 +1,23 @@
...
+@OptIn(ExperimentalTime::class)
+fun <T> Flow<T>.throttleFirst(durationMillis: Long): Flow<T> = flow {
+    var lastEmissionTime: Instant? = null
+
+    collect { value ->
+        val currentTime = Clock.System.now()
+
+        if (lastEmissionTime == null ||
+            (currentTime - lastEmissionTime!!).inWholeMilliseconds >= durationMillis
+        ) {
```
ahmednasserzaza: Using !! may cause a NullPointerException. Please check this sample—I just want to make sure you got the idea.
```
fun <T> Flow<T>.throttleFirst(windowDuration: Long): Flow<T> {
    var job: Job = Job().apply { complete() }
    return onCompletion { job.cancel() }.run {
        flow {
            coroutineScope {
                collect { value ->
                    if (!job.isActive) {
                        emit(value)
                        job = launch { delay(windowDuration) }
                    }
                }
            }
        }
    }
}
```

mohamedd-hassan (approved):

ahmedmahgoub0 (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/base/BaseViewModel.kt
iZyadNasser: @ahmednasserzaza 
Why are they useless? what if we want to make a callback right after the effect finishes?

iZyadNasser (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/base/BaseViewModel.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    private val _effect = MutableSharedFlow<Effect>()
+    val effect = _effect.throttleFirst(300L)
+
+    protected fun updateState(updater: State.() -> State) {
+        _state.update { updater(it) }
+    }
+
+    protected fun sendEffect(
+        event: Effect,
```
ahmednasserzaza: in addition to this is an effect not an event

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/base/BaseViewModel.kt
ahmednasserzaza: @iZyadNasser 
Even if we used to do something like that, this is a special case. We can handle it directly in the ViewModel, so there’s no need to make that function more complex

MahmoodTarek (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/util/extensions.kt
ahmednasserzaza: @ahmeddsamii 
You can use this sample please it's more readable

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/base/BaseViewModel.kt
ahmednasserzaza: @ahmeddsamii

ahmednasserzaza (changes requested):

ahmednasserzaza (approved):

iZyadNasser (approved):

PR #32, opened 2025-09-13
Feature/trn 12 chip component design

skaik-mo:
### How to Use `Tag` Component
  ```
  var selected by remember { mutableStateOf(false) }
    Tag(
        text = "Hello",
        emoji = "\uD83D\uDE00",
        selected = selected,
        onClick = {
            selected = it
        }
    )
```

### This PR also includes a demo video showing the Tag component in action.

https://github.com/user-attachments/assets/bce65374-af29-4b1d-932b-dcfd52c5a865

### CODE REVIEW

**Review by Shahd-Hatem — APPROVED** [review]

File: trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt [hunk 1]
```diff
@@ -0,0 +1,75 @@
...
+import net.thechance.mena.designsystem.presentation.theme.theme.Theme
+import org.jetbrains.compose.ui.tooling.preview.Preview
+
+@Composable
+fun Tag(
+    text: String,
+    emoji: String,
+    selected: Boolean,
+    modifier: Modifier = Modifier,
+    onClick: (Boolean) -> Unit = {}
```
iZyadNasser: In my opinion, I don't like using empty lambdas as a default value for on click functions, because it means that in some cases, there will be a ripple effect when clicking but nothing happens, you have two other options:

- Either make sure if the on click function is required for this component, if it is, then don't give it a default value at all
- If you think the on click is not required and the component can in some cases not be clickable, then make the on click parameter nullable with a null default value, and only put the clickable modifier after checking it is not null, that way you will make sure that there is no unnecessary ripple effect

In our case, I think the first option is best, the Tag component (at least for now) will always be clickable, so better make the on click required

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
```diff
@@ -0,0 +1,75 @@
...
+            .background(color = Theme.colorScheme.primary.onPrimary)
+            .clickable {
+                onClick(!selected)
+            }
+            .padding(10.dp)
+    ) {
+        Row(
+            horizontalArrangement = Arrangement.spacedBy(Theme.spacing._4)
+        ) {
+            Text(
```
fatimafaisalghazi: u could use MenaText here 🥇

ahmeddsamii: Yes and we should delete material 3 from gradle

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
```diff
@@ -0,0 +1,75 @@
...
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.unit.dp
+import net.thechance.mena.designsystem.presentation.theme.theme.Theme
+import org.jetbrains.compose.ui.tooling.preview.Preview
+
+@Composable
+fun Tag(
+    text: String,
+    emoji: String,
+    selected: Boolean,
```
fatimafaisalghazi: is it better to rename it to` isSelected` what do u think ?

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
```diff
@@ -0,0 +1,75 @@
...
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.unit.dp
+import net.thechance.mena.designsystem.presentation.theme.theme.Theme
+import org.jetbrains.compose.ui.tooling.preview.Preview
+
+@Composable
+fun Tag(
```
ahmeddsamii: U can name it CategoryItem, it's more descrptive

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
```diff
@@ -0,0 +1,75 @@
...
+            onClick = {
+                onClick(!selected)
+            },
+        )
+    }
+}
+
+@Preview
+@Composable
+fun TagPreview() {
```
fatimafaisalghazi: u could make it pravite as it is a Preview

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
ahmeddsamii: isSelected

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
```diff
@@ -0,0 +1,75 @@
...
+    val textColor by animateColorAsState(
+        targetValue = if (selected) Theme.colorScheme.shadePrimary else Theme.colorScheme.shadeSecondary
+    )
+    Row(
+        horizontalArrangement = Arrangement.spacedBy(Theme.spacing._8),
+        modifier = modifier
+            .clip(RoundedCornerShape(Theme.radius.full))
+            .background(color = Theme.colorScheme.primary.onPrimary)
+            .clickable {
+                onClick(!selected)
```
iZyadNasser: I think it is unnecessary to give the on click a parameter and pass it from here, the composable that will call this component will provide the state anyways, so it will have access to it from outside, you don't have to pass  it from here, make the component as blind as possible to the state management and logic

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
```diff
@@ -0,0 +1,75 @@
...
+        ) {
+            Text(
+                text = emoji
+            )
+            Text(
+                text = text,
+                color = textColor
+            )
+        }
+        net.thechance.mena.designsystem.presentation.component.button.radioButton.RadioButton(
```
iZyadNasser: Can't you import this instead of calling the whole path here?

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
iZyadNasser: This is a minor detail but it's better to prefix parameters of Boolean values with "is", so "isSelected" instead of "selected"

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
fatimafaisalghazi: can we add this in the import section in the top , is it work ?

iZyadNasser (commented):
Really great work Mohamed, only some minor comments

fatimafaisalghazi (approved):
great work 💯

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
skaik-mo: Yes, I think this makes it more readable.

skaik-mo: Yes, I think this makes it more readable.

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
skaik-mo: Yes, the clickable is required, so I’ll update it.

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
skaik-mo: I didn’t notice it in the design. I’ll check the design system and use it.

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
skaik-mo: It’s called `Tag` in the design. What do you think about instead using the name `CategoryItem` for the component, @iZyadNasser?

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
skaik-mo: Yes, you’re right, it’s unnecessary. I’ll remove it.

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
```diff
@@ -0,0 +1,74 @@
...
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.unit.dp
+import net.thechance.mena.designsystem.presentation.component.button.radioButton.RadioButton
+import net.thechance.mena.designsystem.presentation.component.text.MenaText
+import net.thechance.mena.designsystem.presentation.theme.theme.Theme
+import org.jetbrains.compose.ui.tooling.preview.Preview
+
+@Composable
+fun Tag(
+    text: String,
```
ahmednasserzaza: We need two things:

First: Create CategoryUiState as a shared UiState with its mapper from entity to UiState, then pass the object directly from CategoryUiState. This way, we don’t need to use text, emoji, or any other category-related values as separate parameters.

Second: Add a shared data class called Selectable, which will help us with any selectable component, not just this one.

You can check this, and of course, it should be shared.
```
data class Selectable<T>(
    val value: T,
    val isSelected: Boolean = false,
)
```
 in addition we need at least to pass categoryId onClick
 So I am expecting the function parameters to be like that 
 ```
@Composable
 fun CategoryItem(
 category:Selectable<CategoryUiState>,
 modifier: Modifier = Modifier,
 onClick: (Int) -> Unit
 ){}
```

File: trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt [hunk 9]
```diff
@@ -0,0 +1,74 @@
...
+) {
+    val textColor by animateColorAsState(
+        targetValue = if (isSelected) Theme.colorScheme.shadePrimary else Theme.colorScheme.shadeSecondary
+    )
+    Row(
+        horizontalArrangement = Arrangement.spacedBy(Theme.spacing._8),
+        modifier = modifier
+            .clip(RoundedCornerShape(Theme.radius.full))
+            .background(color = Theme.colorScheme.primary.onPrimary)
+            .clickable { onClick() }
```
ahmednasserzaza: please check if it has a ripple effect then disable it by making             
indication = null,

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/Tag.kt
ahmednasserzaza: I agree with @ahmeddsamii 
@skaik-mo please rename it

ahmednasserzaza (changes requested):

ahmedmahgoub0 (approved):

iZyadNasser (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/CategoryItem.kt
```diff
@@ -0,0 +1,100 @@
...
+        )
+        RadioButton(
+            isSelected = category.isSelected,
+            onClick = { onClick(category.value.id) },
+        )
+    }
+}
+
+@Composable
+private fun CategoryNameAndEmoji(
```
ahmednasserzaza: It's not necessary to separate it into a component, but if you’d like to, don’t pass textColor as a parameter—just handle its state animation inside it directly

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/model/CategoryUiState.kt
```diff
@@ -0,0 +1,7 @@
+package net.thechance.mena.trends.presentation.shared.model
+
+data class CategoryUiState(
```
ahmednasserzaza, line 3: all its properties should have a default value, regarding id make it nullable with null as a default value

ahmednasserzaza (changes requested):

ahmednasserzaza (approved):
Well done bro ❤️

PR #37, opened 2025-09-13
feature/DKN-9-update-add-dukan-button-to-reflect-pending-status

ahmedNaser7:
The Pr Description : 
- add `TopAppBar` component of `Dukan Main Screen`
- handle the state of `DukanButton` while the dukan is pending or not
- make `DukanButton` in the `TopAppBar` animates its visibility based on the `isDukanPending` state.
- add initial classes of `Dukan Main Screen` :
    - `MainScreen`
    - `MainViewMode`
    - `MainScreanUiState`
    - `MainEffect`
    - `MainInteraticonListener`
 - test the `MainViewModel`with 100 precent coverage


dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeScreen.kt
```diff
@@ -0,0 +1,68 @@
...
+    state: HomeScreenUiState
+) {
+    Column(
+        modifier = Modifier
+            .background(color = Theme.colorScheme.background.surface)
+            .statusBarsPadding(),
+        verticalArrangement = Arrangement.Center
+    ) {
+        TopAppBar(
+            modifier = Modifier,
```
YoussefmSaber: There is no need to send the modifier if it do nothing

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeScreen.kt
```diff
@@ -0,0 +1,68 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import net.thechance.mena.designsystem.presentation.theme.theme.MenaTheme
+import net.thechance.mena.designsystem.presentation.theme.theme.Theme
+import net.thechance.mena.dukan.presentation.screen.home.components.TopAppBar
+import org.jetbrains.compose.ui.tooling.preview.Preview
+
+@Composable
+fun HomeScreen() {
```
haideraljawasim: I see the `MainScreen`  name more clear

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeScreen.kt
```diff
@@ -0,0 +1,68 @@
...
+            isDukanPending = state.isPending
+        )
+
+        Spacer(modifier = Modifier.weight(1f))
+    }
+}
+
+@Preview
+@Composable
+fun HomeScreenPreview() {
```
haideraljawasim: you can make it private

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeScreenUiState.kt
```diff
@@ -0,0 +1,6 @@
+package net.thechance.mena.dukan.presentation.screen.home
+
+data class HomeScreenUiState(
+    val error: String? = null,
```
haideraljawasim: will we use this error message ? i think we  should place here snack bar data class,  so like can use it for example if the operation was success or failed add to the message

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeViewModel.kt
```diff
@@ -0,0 +1,33 @@
...
+                    copy(isPending = isDukanExist)
+                }
+            },
+            onError = {
+                updateState {
+                    copy(error = it.message)
+                }
+            },
+        )
+    }
```
YoussefmSaber: it's better to split the content of this function to multiple smaller functions and use call by reference for it,
so the function should look something like this.
```
private fun getDukanButtonState() {
        tryToExecute(
            block =  dukanRepository::isUserHasDukan,
            onSuccess = ::onGetDukanButtonStateSuccess,
            onError = ::onGetDukanButtonStateError
        )
    }
        
private fun onGetDukanButtonStateSuccess(isDukanExist: Boolean) {
// Content of the onSuccess
}

private fun onGetDukanButtonStateError(throwable: Throwable) {
// Content of the onError
}
```
and the name of the function can be enhanced, instead of `getDukanButtonState` it should be `getDukanState`, the reason for this we shouldn't force the usage of the state to the ui, maybe today we use the state to change an icon, maybe tomorrow we use it to display a message, that's why we should not relate a state with a ui element

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeScreenUiState.kt
```diff
@@ -0,0 +1,6 @@
...
+    val isPending: Boolean = false
```
haideraljawasim: I know the task is about pending dukan but  i see naming it now `isUserHasDukan` more clear since the pending screen later will be user dukan screen , I'm just suggestion

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/components/TopAppBar.kt
```diff
@@ -0,0 +1,108 @@
...
+                                easing = EaseIn
+                            )
+                        ) togetherWith(fadeOut(
+                            animationSpec = tween(
+                                durationMillis = 500,
+                                delayMillis = 100,
+                                easing = EaseOut
+                            ))
+                        )
+                    },
```
haideraljawasim: i suggest that we should have like a file to store like these animations so we use the same spec across the feature

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeInteractionListener.kt
```diff
@@ -0,0 +1,5 @@
+package net.thechance.mena.dukan.presentation.screen.home
+
+interface HomeInteractionListener {
+    fun onAddDukanButtonClicked()
```
Mohamed-Ashraf-2210: I think onAddDukanClicked is better

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/components/TopAppBar.kt
```diff
@@ -0,0 +1,108 @@
...
+                    label = stringResource(resource = Res.string.dukan_button)
+                )
+                { isPending ->
+                    Box(
+                        modifier = Modifier
+                            .size(40.dp)
+                            .background(color = Theme.colorScheme.background.surfaceLow, shape = RoundedCornerShape(Theme.spacing._12))
+                            .clip(shape = RoundedCornerShape(Theme.spacing._12))
```
haideraljawasim: are't there radius values in design system , why you used here spacing value for radius

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeScreenUiState.kt
ahmedNaser7: No, the design doesn't mention using a snackbar, and the message is only for handling the error and keeping it until the requirements of the UI change, and also that is what is required from me in this task.

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeEffect.kt
```diff
@@ -0,0 +1,6 @@
+package net.thechance.mena.dukan.presentation.screen.home
+
+sealed class HomeEffect {
+    object NavigateToAddDukanScreen : HomeEffect()
```
Mohamed-Ashraf-2210: i think this name without Screen is better
`NavigateToAddDukan`

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/components/TopAppBar.kt
```diff
@@ -0,0 +1,108 @@
...
+                    }
+                }
+            }
+        }
+    )
+}
+
+@Preview
+@Composable
+fun TopAppBarPreview() {
```
haideraljawasim: same this should be private function no need to leave public

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeViewModel.kt
```diff
@@ -0,0 +1,33 @@
...
+
+class HomeViewModel(
+    private val dukanRepository: DukanRepository
+) : BaseViewModel<HomeScreenUiState, HomeEffect>(HomeScreenUiState()), HomeInteractionListener {
+
+    init {
+        getDukanButtonState()
+    }
+
+    private fun getDukanButtonState() {
```
Mohamed-Ashraf-2210: i think `getIsUserHasDukan` will be better

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeScreenUiState.kt
ahmedNaser7: okay , it's all up to business requirements , later we can update the naming

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeViewModel.kt
```diff
@@ -0,0 +1,33 @@
...
+        tryToExecute(
+            block = { dukanRepository.isUserHasDukan() },
+            onSuccess = { isDukanExist ->
+                updateState {
```
Mohamed-Ashraf-2210: can separate this to another function like `getIsUserHasDukanSucces`

Mohamed-Ashraf-2210 (changes requested):

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/components/TopAppBar.kt
```diff
@@ -0,0 +1,108 @@
...
+import net.thechance.mena.designsystem.presentation.theme.theme.Theme
+import org.jetbrains.compose.resources.painterResource
+import org.jetbrains.compose.resources.stringResource
+import org.jetbrains.compose.ui.tooling.preview.Preview
+
+@Composable
+fun TopAppBar(
+    modifier: Modifier = Modifier,
+    onAddDukanButtonClicked: () -> Unit = {},
+    isDukanPending: Boolean
```
haideraljawasim: the optional parameters  should be last thing

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/components/TopAppBar.kt
ahmedNaser7: I think it's more flexible to use each custom animation in its place.

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/components/TopAppBar.kt
```diff
@@ -0,0 +1,108 @@
...
+                            )
+                        } else {
+                            Icon(
+                                painter = painterResource(resource = Res.drawable.ic_add_dukan),
+                                contentDescription = stringResource(resource = Res.string.add_dukan_icon)
+                            )
+                        }
+                    }
+                }
+            }
```
YoussefmSaber: There is 2 questions here, 
- why did you wrap `AnimatedContent` with a `row` when there is no item's next to it.
- why didn't you extract the animations from `transitionSpec` and use it as a variable.

how can we fix this code and make it much better and easier to read.
- `First`: you are going to remove the `Row` as it don't add any value to the `composable`.
- `Second`: you will extract the `transitionSpec` content and make them variables, to be like this: 
```
val fadeInAnimation = fadeIn(
        animationSpec = tween(
        durationMillis = 500,
        delayMillis = 100,
        easing = EaseIn
    )
)

val fadeOutAnimation = fadeOut(
        animationSpec = tween(
        durationMillis = 500,
        delayMillis = 100,
        easing = EaseOut
    )
)
 ```
 then your transitionSpec look like this:
 ```
 transitionSpec = { fadeInAnimation togetherWith fadeOutAnimation }
```
- `Third`: the icon section you have 2 option to apply, either you add the statement to the `drawable` or extract it to a `private composable` and send to it the `isPending` this will help the readability for the code.
-  and `Finally`:  extract all the content of `trailingContent` to a `private composable` and call it in here.

And i hope you take this rule, it's better to not exceed the 3 indentation tabs when creating a composable, as this will effect the reading of the composable.

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/components/TopAppBar.kt
ahmedNaser7: Thanks for noticing this missing; I will replace it with radius values of the design system.

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/components/TopAppBar.kt
haideraljawasim: Im just answering for the why there's  `Row`   , `AnimatedContent` can  be only  inside row or column scope

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeScreenUiState.kt
haideraljawasim: handling error for what will i display it ? no , so i believe we shouldn't keep as it useless literally

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeInteractionListener.kt
ahmedNaser7: why its better ? , i try to make the name more clear and the user will click on button

Mohamed-Ashraf-2210 (commented):
Great Job

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeScreen.kt
ahmedNaser7: np , i  will delete it

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeViewModel.kt
ahmedNaser7: I agree with the point of using getDukanState, but why do we separate the function into another small function? 

What makes it better?

 In my opinion, it's over-engineering to make that while I have a small code and a small function.

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/components/TopAppBar.kt
YoussefmSaber: @haideraljawasim  the `animatedContent` can be used without any scope.

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeViewModel.kt
YoussefmSaber: it's not an over-engineering, it's just to make the code base more consistent, the function it's self looks ugly, i don't mean that your code is ugly, but the implementation for the function it's self, why is that, the function content it's self exceeds a 3 indentation mark. ok and what if we want to add more check to the on success function and on error, the function will increase in size, it's small tweak but it will improve the readability more and can help with testing the code.

ahmedNaser7: i agree with your idea in case we have a functions include a large amounts of logic code . 
but its not big deal i will see what is the best for make it more readability and use it .
thanks for your suggestions ❤️

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/components/TopAppBar.kt
ahmedNaser7: the answer for questions 1 : its only way that it work with me to use animatedContent but i will try to pass it as composable higher function 
the answer for questions 2 : it's all about reusability , there is many way to handle transitionSpec , like make it a private function and pass it like that 

```
 private fun fadeTransitionSpec(): ContentTransform {
    return fadeIn(
        animationSpec = tween(
            durationMillis = 500,
            delayMillis = 100,
            easing = EaseIn
        )
    ) togetherWith fadeOut(
        animationSpec = tween(
            durationMillis = 500,
            delayMillis = 100,
            easing = EaseOut
        )
    )
}  
```

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeScreenUiState.kt
ahmedNaser7: i agree with your opinion that is not required and never use this error message . but still need to handle the exception while user using the screen .
@mostafa-n3ma i need your opinion in this point , what is the best here ?

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/components/TopAppBar.kt
```diff
@@ -0,0 +1,108 @@
...
+                    Box(
+                        modifier = Modifier
+                            .size(40.dp)
+                            .background(color = Theme.colorScheme.background.surfaceLow, shape = RoundedCornerShape(Theme.spacing._12))
+                            .clip(shape = RoundedCornerShape(Theme.spacing._12))
+                            .clickable(onClick = onAddDukanButtonClicked),
+                        contentAlignment = Alignment.Center
+                    ) {
+                        if (isPending) {
+                            Icon(
```
Ahmed-AlMancy: Try using the MenaIcon from design system not material 3, the same should go for the MenaText

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeScreen.kt
```diff
@@ -0,0 +1,68 @@
...
+    MenaTheme {
+        Box(
+            modifier = Modifier
+                .fillMaxWidth()
+                .background(Theme.colorScheme.background.surface),
+            contentAlignment = Alignment.Center
+        ) {
+            HomeContent(
+                homeInteractionListener = object : HomeInteractionListener {
```
Ahmed-AlMancy: what do you think about moving this object from the preview because in the future the HomeIneteraction listener will have a lot of functions to be implemented, and this way this preview will have like lots of code just to preview, so you might create a stubPreview package in the util and create an object there will will hold the interaction listener functions of HomeInteractionListener, i think it will be better, what do you think?

ahmedNaser7: Thanks for sharing your thoughts. , but once we implement our DI module or package ( Koin ) ,  i will replace HomeContent() with HomeScreen() so i will didnt need to pass any parameter .

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeEffect.kt
ahmedNaser7: i think its clear when i use `Screen` at the end

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeScreenUiState.kt
mostafa-n3ma: I think Haider has a point here but to gave my opinion :
ok add (errorMessage) saying "something wrong happened" just for now and then when we implement a snackBar or any warning approach with localized strings we will update it relatively 

you can also add two properties like 
userHasDukan:Boolean
dukanStatus:Status

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/components/TopAppBar.kt
mostafa-n3ma: animation is a plus here not really required in the design so I see  both of your suggestions are correct

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeInteractionListener.kt
mostafa-n3ma: I think onDukanButtonClicked() is better since we will update it between two status (None & pending) with two navigation effects and two icon cases (add & pending dukan) not just adding

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeViewModel.kt
mostafa-n3ma: I think you two missed the main Idea here !!
- we should first check if the user has a dukan or not  (here we decide to navigate to  creatDukanScreen  or not and show createDukan Icon )
- and then check the dukan status if pending or not to choose rather we navigate to pending screen or not and show Dukan Icon 
 you can leave the other dukan staus cases for latter 
 @YoussefmSaber 
 @ahmedNaser7

mostafa-n3ma (commented):
please consider the valuable comments and notes from our team mates 
and update accordingly

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeViewModel.kt
ahmedNaser7: you are correct , but The task is about updating the state of the Dukan icon, like changing from the "create Dukan icon" to the "Dukan icon (pending)."
But the navigation is not included in the task requirements , so I didn't make implementation of dukanButtonNavigation
@mostafa-n3ma

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/HomeScreenUiState.kt
ahmedNaser7: 1 - I think I cannot add an error message until the business requirement changes; we use only the onSuccess scope function. What is your opinion about that ?
```
private fun getDukanState() {
        tryToExecute(
            block = dukanRepository::isUserHasDukan,
            onSuccess = ::onGetDukanStateSuccess,
        )
    }

private fun onGetDukanStateSuccess(isDukanExist: Boolean) {
        updateState {
            copy(isUserHasDukan = isDukanExist)
        }
    }
```
2 - already change isPending to `isUserHasDukan:Boolean` in last commit 
3 - why dukanStatus ? and we only want to know if Request of Dukan is Pending or not , and if i add it should be like that 
```
enum class DukanStatus {
 PENDING,NEW
}
```

ahmedNaser7:
> please consider the valuable comments and notes from our team mates and update accordingly

done ✅

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/MainScreen.kt
```diff
@@ -0,0 +1,67 @@
...
+    val state = viewModel.state.collectAsStateWithLifecycle()
+    MainContent(
+        mainInteractionListener = viewModel,
+        state = state.value
+    )
+}
+
+@Composable
+private fun MainContent(
+    mainInteractionListener: MainInteractionListener,
```
YoussefmSaber: it will be better to call it just `listener` to be consistent with the code base.

YoussefmSaber (approved):

Mohamed-Ashraf-2210 (approved):

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/MainViewModel.kt
```diff
@@ -0,0 +1,30 @@
...
+    private fun getDukanState() {
+        tryToExecute(
+            block = dukanRepository::isUserHasDukan,
+            onSuccess = ::onGetDukanStateSuccess,
+        )
+    }
+
+    private fun onGetDukanStateSuccess(isDukanExist: Boolean) {
+        updateState {
+            copy(isUserHasDukan = isDukanExist)
```
mostafa-n3ma: we should check if the user has dukan or not and then if the dukan status is pending or not

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/MainScreen.kt
```diff
@@ -0,0 +1,67 @@
...
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import net.thechance.mena.designsystem.presentation.theme.theme.MenaTheme
+import net.thechance.mena.designsystem.presentation.theme.theme.Theme
+import net.thechance.mena.dukan.presentation.screen.home.components.TopAppBar
+import org.jetbrains.compose.ui.tooling.preview.Preview
+
+@Composable
+fun MainScreen() {
+    val viewModel = MainViewModel(dukanRepository = TODO(reason = "should using koin di"))
+    val state = viewModel.state.collectAsStateWithLifecycle()
```
mostafa-n3ma, line 23: you have two effects and yet didn't handle them here please collect the latest effect and should have blocks{ } to handle them with the relative navigation action

mostafa-n3ma (changes requested):

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/Mapper.kt
```diff
@@ -0,0 +1,10 @@
+package net.thechance.mena.dukan.presentation.screen.home
+
+import net.thechance.mena.dukan.domain.entity.Dukan
+
+fun Dukan.Status.toUiState(): MainScreenUiState.DukanStatusUi{
+    return when(this){
+        Dukan.Status.Pending -> MainScreenUiState.DukanStatusUi.Pending
```
AbdelrhmanAyman20470, line 7: i think better to use status directly

ahmedNaser7: Make the each screen depend on his  uiState is recommended for separation between layers

AbdelrhmanAyman20470: maybe u didnt understand me right i mean import Status and remove prefix dukan
like that 
```
import net.thechance.mena.dukan.domain.entity.Dukan.Status

fun Status.toUiState(): MainScreenUiState.DukanStatusUi {
    return when (this) {
        Status.Pending -> MainScreenUiState.DukanStatusUi.Pending
        Status.None -> MainScreenUiState.DukanStatusUi.None
    }
}
```

ahmedNaser7: Now i understand , but what is the reason  ?

dukan_presentation/src/commonTest/kotlin/net/thechance/mena/dukan/presentation/screen/home/MainViewModelTest.kt
```diff
@@ -0,0 +1,148 @@
...
+            mainViewModel.state.test {
+                awaitItem()
+                val secondEmit = awaitItem()
+                assertEquals("Dukan not found", secondEmit.errorMessage)
+                cancelAndIgnoreRemainingEvents()
+            }
+        }
+
+    @Test
+    fun `When the user doesn't have Dukan and clicks on the Dukan button, then it should emit NavigateToAddDukanScreen`() =
```
mostafa-n3ma: for some reason this characters "," is illegal in the test function name and throwing error when running it on IosSimulatorArm64 my be du to some mis configuration on with iOS  but removing them fixing the problem and get more readable

mostafa-n3ma (approved):
you tests looks solid , thank you Ahmed

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/home/Mapper.kt
AbdelrhmanAyman20470: better for readability

PR #46, opened 2025-09-14
Feature/TRN-23 setup ktor http client for trends

ahmedmahgoub0:
### Description

Setup `ktor` client for trends 

---

**Added**  
- `ktor-client-core`
- `ktor-content-negotiation`, 
- `ktor-serialization-json`, 
- `ktor-client-logging`
- `ktor-client-cio` for Android
- `ktor-client-darwin` for IOS

**Chnaged**
- Added the above dependencies to the `TrendsDataModule`'s `build.gradle` file
- Provided with koin `NetworkModule` and included it in the `TrendsDataModule`

### DIFF
--- a/trends_data/build.gradle.kts
+++ b/trends_data/build.gradle.kts
@@ -20,7 +20,7 @@ kotlin {
 
     sourceSets {
         androidMain.dependencies {
-            implementation(libs.ktor.client.okhttp)
+
         }
         commonMain.dependencies {
             implementation(projects.trendsDomain)
@@ -30,7 +30,7 @@ kotlin {
             implementation(libs.bundles.ktor)
         }
         iosMain.dependencies {
-            implementation(libs.ktor.client.darwin)
+
         }
     }
 
--- a/trends_data/build.gradle.kts
+++ b/trends_data/build.gradle.kts
@@ -6,6 +6,7 @@ plugins {
     alias(libs.plugins.kotlinMultiplatform)
     alias(libs.plugins.androidLibrary)
     alias(libs.plugins.ksp)
+    alias(libs.plugins.kotlinx.serialization)
 }
 
 kotlin {
@@ -19,16 +20,17 @@ kotlin {
 
     sourceSets {
         androidMain.dependencies {
-
+            implementation(libs.ktor.client.okhttp)
         }
         commonMain.dependencies {
             implementation(projects.trendsDomain)
             implementation(libs.koin.core)
             api(libs.koin.annotations)
             implementation(libs.kotlinx.datetime)
+            implementation(libs.bundles.ktor)
         }
         iosMain.dependencies {
-
+            implementation(libs.ktor.client.darwin)
         }
     }
 
--- a/trends_data/build.gradle.kts
+++ b/trends_data/build.gradle.kts
@@ -20,7 +20,7 @@ kotlin {
 
     sourceSets {
         androidMain.dependencies {
-
+            implementation(libs.ktor.client.cio)
         }
         commonMain.dependencies {
             implementation(projects.trendsDomain)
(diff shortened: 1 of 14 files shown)


trends_data/build.gradle.kts
```diff
@@ -19,16 +20,17 @@ kotlin {
 
     sourceSets {
         androidMain.dependencies {
-
+            implementation(libs.ktor.client.okhttp)
```
ahmednasserzaza: You can use CIO engine instead, I think its more stable than okhttp

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/di/NetworkModule.kt
```diff
@@ -0,0 +1,58 @@
...
+import io.ktor.client.plugins.defaultRequest
+import io.ktor.client.plugins.logging.LogLevel
+import io.ktor.client.plugins.logging.Logger
+import io.ktor.client.plugins.logging.Logging
+import io.ktor.serialization.kotlinx.json.json
+import kotlinx.serialization.json.Json
+import org.koin.core.annotation.Module
+import org.koin.core.annotation.Single
+
+expect fun createHttpEngine(): HttpClientEngine
```
ahmednasserzaza: No need for that functionality, KMP by default detacts the current running platform and uses their engine, if it has been added to androidMain and iosMain, So you can remove all those functionality related to HttpClientEngine

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/di/NetworkModule.kt
```diff
@@ -0,0 +1,58 @@
...
+
+@Module
+class NetworkModule {
+    @Single
+    fun provideHttpClient(engine: HttpClientEngine): HttpClient {
+        return HttpClient(engine) {
```
ahmednasserzaza: don't pass engine here, it will detect it automatically

ahmednasserzaza (changes requested):

trends_data/build.gradle.kts
ahmedmahgoub0: @ahmednasserzaza ok, I will change it, for learning, could you share with me  what makes it more stable or better decision here ?

ahmednasserzaza: It doesn't matter for now if we used CIO or okHttp, I prefere CIO becuase it has been written by Ktor team itself, and we are using ktor as a network client So, I expected less issues than using another engine

ahmednasserzaza (approved):
Well done

mohamedd-hassan (approved):

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/di/NetworkModule.kt
```diff
@@ -0,0 +1,52 @@
...
+        return HttpClient {
+            defaultRequest {
+                // TODO: add base url
+            }
+
+            install(Logging) {
+                level = LogLevel.ALL
+                logger = object : Logger {
+                    override fun log(message: String) {
+                        println("Http client: $message")
```
iZyadNasser, line 28: Maybe consider the "Http client" a TAG and put it as a const in companion object

private const val LOG_TAG = "Http client"

iZyadNasser (approved):
Great Job

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/di/NetworkModule.kt
ahmedmahgoub0: I think it's just for clarification and to add to the message, no need to do that?

PR #67, opened 2025-09-15
Feature/WLT-6 Wallet Screen UI

MuhammedEdrees:
### PR Content:
- Added simple `BaseViewModel`.
- Implemented `WalletScreen` with `BalanceCard`.

| Wallet Screen | Balance Card | Error Snackbar |
|--------|-------|------- |
| image | image|image |


wallet_presentation/src/commonMain/composeResources/values/strings.xml
Al-Qassim, line 1: add Arabic trasulation

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/utils/ObserveAsEffect.kt
```diff
@@ -0,0 +1,25 @@
...
+import androidx.lifecycle.compose.LocalLifecycleOwner
+import androidx.lifecycle.repeatOnLifecycle
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.withContext
+
+@Composable
+fun <T> ObserveAsEffect(
+    effect: Flow<T>,
+    onEvent: (T) -> Unit,
```
Al-Qassim: make `onEvent` function suspended
`onEvent: suspended (T) -> Unit`
so we can use `delay` function inside it if we want, or and other suspended function

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/utils/StringUtils.kt
```diff
@@ -0,0 +1,16 @@
...
+    val wholePart = balance.toLong()
+    val decimalPart = ((abs(balance) % 1) * 100).toLong()
+
+    val wholeString = wholePart.toString()
+        .reversed()
+        .chunked(3)
+        .joinToString(",")
+        .reversed()
+
+    return "$wholeString.${decimalPart.toString().padStart(2, '0')}"
```
Al-Qassim: well done, and I think we can make it clearer

```
fun formatBalance(balance: Double): String {
    val integerPart = balance.toLong()
    val decimalPart = ((abs(balance) % 1) * 100).toLong()

    val integerString = wholePart
        .toString()
        .reversed()
        .chunked(3)
        .joinToString(",")
        .reversed()

    val decimalString = decimalPart
        .toString()
        .padStart(2, '0')
    
    return "$wholeString.$decimalString "
```

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/base/BaseViewModel.kt
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    protected fun <T> tryToExecute(
+        callee: suspend () -> T,
+        onSuccess: ((T) -> Unit)? = null,
+        onError: ((Throwable) -> Unit)? = null,
+        onStart: (suspend () -> Unit)? = null,
+        onFinish: (() -> Unit)? = null,
+        dispatcher: CoroutineDispatcher = Dispatchers.IO,
```
Al-Qassim: all function can be made suspended, this will allow us to take full advantage of coroutines 
```
protected fun <T> tryToExecute(
        callee: suspend () -> T,
        onSuccess: (suspend (T) -> Unit)? = null,
        onError: (suspend (Throwable) -> Unit)? = null,
        onStart: (suspend () -> Unit)? = null,
        onFinish: (suspend () -> Unit)? = null,
        dispatcher: CoroutineDispatcher = Dispatchers.IO,
    )
```

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/wallet/component/BalanceCard.kt
Al-Qassim, line 1: the balance should not appear until the request finishes and the balance is fetched. 
 
During that time, we should should 3 dots instead of the balance

https://github.com/user-attachments/assets/ddc91603-6e6e-4419-9b04-c44fde03db21

Al-Qassim: also, there is a failed state now 

image

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/wallet/component/BalanceCard.kt
```diff
@@ -0,0 +1,155 @@
...
+                shape = RoundedCornerShape(Theme.radius.xl)
+            )
+            .clip(RoundedCornerShape(Theme.radius.xl))
+            .then(modifier)
+    ) {
+        val parentWidthPx = with(LocalDensity.current) { maxWidth.toPx() }
+
+        AnimatedCoinImage(
+            modifier = Modifier
+                .padding(top = 12.dp)
```
Al-Qassim, line 64: it's a little bit weird that you put 12 here and 8 below. why not put the whole 20 below and remove this 12?

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/wallet/component/TopNotchShape.kt
Al-Qassim, line 1: ok, this shape class gets most of the job done.
but it can be improved

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/wallet/component/TopNotchShape.kt
```diff
@@ -0,0 +1,126 @@
...
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.LayoutDirection
+import androidx.compose.ui.unit.dp
+
+class TopNotchShape(
+    private val offset: Float,
+    private val circleRadius: Dp,
+    private val cornerRadius: Dp,
+    private val circleGap: Dp = 5.dp,
+) : Shape {
```
Al-Qassim: the parameters `circleRadius` at line 15 is misleading, the actual shape of the cutout in the design is not a circle, and in your code you didn't use a circle or arc for the cutout, you actually used a cubic shape (which is very correct).

Al-Qassim: also the parameter `circleGap` is totally unnecessary, it doesn't reflect anything in the design. and it makes the cutout a little bit wider

I suggest removing it

Al-Qassim: I see that a better parameters for this class is the cutout width and depth, which are 100dp and 32dp respectively

image

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/wallet/component/TopNotchShape.kt
```diff
@@ -0,0 +1,126 @@
...
+
+    private fun getPath(size: Size, density: Density): Path {
+        val cutoutCenterX = offset
+        val cutoutRadius = density.run { (circleRadius + circleGap).toPx() }
+        val cornerRadiusPx = density.run { cornerRadius.toPx() }
+        val cornerDiameter = cornerRadiusPx * 2
+        return Path().apply {
+            val cutoutEdgeOffset = cutoutRadius * 2.5f
+            val cutoutLeftX = cutoutCenterX - cutoutEdgeOffset
+            val cutoutRightX = cutoutCenterX + cutoutEdgeOffset
```
Al-Qassim: you may want to make  
`cutoutEdgeOffset = cutoutWidth / 2 + 12`
instead of multiplying with the arbitrary number of 2.5

image

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/wallet/component/TopNotchShape.kt
```diff
@@ -0,0 +1,126 @@
...
+    private fun Path.drawTopLeftEdge(
+        cutoutLeftX: Float,
+        cornerRadiusPx: Float,
+        cornerDiameter: Float
+    ) {
+        val realLeftCornerDiameter = if (cutoutLeftX >= cornerRadiusPx) {
+            cornerDiameter
+        } else {
+            cutoutLeftX * 2
+        }
```
Al-Qassim: always try to reduce the number of parameters in a function, for example, you can remove the `cornerDiameter` parameter, since you can compute it from `cornerRadiusPx * 2`

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/wallet/component/TopNotchShape.kt
```diff
@@ -0,0 +1,126 @@
...
+        )
+        cubicTo(
+            x1 = cutoutCenterX + cutoutRadius,
+            y1 = cutoutRadius,
+            x2 = cutoutCenterX + cutoutRadius,
+            y2 = 0f,
+            x3 = cutoutRightX,
+            y3 = 0f,
+        )
+    }
```
Al-Qassim: I think you miss a parameter that we can call `cutoutRoundnessControl`, it can be a parameter in the class constructor, it can have a value of 12 or 20 for example, and I see that it will be used as following in the code
```
        cubicTo(
            x1 = cutoutLeftX + cutoutRoundnessControl,
            y1 = 0f,
            x2 = cutoutCenterX - cutoutRoundnessControl,
            y2 = cutoutDepth,
            x3 = cutoutCenterX,
            y3 = cutoutDepth,
        )
        cubicTo(
            x1 = cutoutCenterX + cutoutRoundnessControl,
            y1 = cutoutDepth,
            x2 = cutoutRightX - cutoutRoundnessControl,
            y2 = 0f,
            x3 = cutoutRightX,
            y3 = 0f,
        )
    
```

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/wallet/WalletScreen.kt
```diff
@@ -0,0 +1,133 @@
...
+@Composable
+private fun WalletContent(
+    state: WalletUiState,
+    interactionListener: WalletInteractionListener
+) {
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+            .statusBarsPadding()
+            .background(Theme.colorScheme.background.surface)
```
Al-Qassim: apply the `background` modifier before the `statusBarsPadding`

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/wallet/WalletScreen.kt
```diff
@@ -0,0 +1,133 @@
...
+        AnimatedContent(
+            targetState = state.isLoading,
+            modifier = Modifier.weight(1f)
+        ) { isLoading ->
+            if (isLoading) {
+                LoadingContent()
+            } else {
+                MainContent(state = state, interactionListener = interactionListener)
+            }
+        }
```
Al-Qassim: 1. cross fade will better
2. not the whole screen should disappear when loading. only the balance card have a loading, error, and success states. the rest of the screen should appear as normal, even in loading and error

MuhammedEdrees: This loading was before the new loading dots was added to the card. I'll update it with the new design.

gehadepada:
i think you should provide multiple sizes for this image "img_silver" for better performance on different devices

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/wallet/component/BalanceCard.kt
```diff
@@ -0,0 +1,155 @@
...
+import org.jetbrains.compose.resources.stringResource
+import org.jetbrains.compose.ui.tooling.preview.Preview
+
+@Composable
+fun BalanceCard(
+    balance: Double,
+    modifier: Modifier = Modifier,
+) {
+    BoxWithConstraints(
+        modifier = Modifier
```
gehadepada, line 71: i think  BoxWithConstraints should take modifier as a parameter that passed to BalanceCard instead of using new modifier ??

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/utils/ObserveAsEffect.kt
```diff
@@ -0,0 +1,25 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.withContext
+
+@Composable
+fun <T> ObserveAsEffect(
+    effect: Flow<T>,
+    onEffect: suspend (T) -> Unit,
+) {
+    val lifecycleOwner = LocalLifecycleOwner.current
```
gehadepada, line 17: it's better to for memory to use  LocalLifecycleOwner.current instead of declaring new variable if you can use it directly inside lunched effect

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/wallet/component/BalanceCard.kt
MuhammedEdrees: The passed modifier paramter is indeed used at the end of modifier chain with `then`.

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/utils/ObserveAsEffect.kt
MuhammedEdrees: `LocalLifecycleOwner` is a composition local which means it cannot be called inside `LaunchedEffect` + it's used in two different places so i think its fine to remain as it is.

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/wallet/component/TopNotchShape.kt
MuhammedEdrees: Hmm, in this case i think a multiplier would make more sense to scale better with cutoutDepth without the need to adjust both separately.

Al-Qassim: ok, also a good idea

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/wallet/component/TopNotchShape.kt
```diff
@@ -0,0 +1,160 @@
...
+
+    private fun Path.drawCutout(
+        cutoutCenterX: Float,
+        cutoutDepth: Float,
+        cutoutRightX: Float
+    ) {
+        val controlPointOffset = cutoutDepth * cutoutRoundness
+
+        cubicTo(
+            x1 = cutoutCenterX - controlPointOffset,
```
Al-Qassim: this should be `x1 = cutoutLeftX + controlPointOffset`

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/wallet/component/TopNotchShape.kt
```diff
@@ -0,0 +1,160 @@
...
+            y1 = 0f,
+            x2 = cutoutCenterX - controlPointOffset,
+            y2 = cutoutDepth,
+            x3 = cutoutCenterX,
+            y3 = cutoutDepth,
+        )
+        cubicTo(
+            x1 = cutoutCenterX + controlPointOffset,
+            y1 = cutoutDepth,
+            x2 = cutoutCenterX + controlPointOffset,
```
Al-Qassim: this should be `x2 = cutoutRightX - controlPointOffset`

Al-Qassim (approved):

PR #71, opened 2025-09-16
Feature/IDN 15 authentication domain

poula22:
- Add authentication Repository and login use case.
- Implement mobile number validation for MENA countries
- Add custom domain exceptions
- Add test cases for all MENA countries mobile numbers and for all other login validations


identity_domain/src/jvmTest/kotlin/net/thechance/mena/identity/domain/useCase/LoginUseCaseTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    fun `should return true when password length greater than 8`() {
+        val password = "password1"
+
+        val isValid = loginUseCase.isPasswordValid(password)
+
+        assertThat(isValid).isTrue()
+    }
+
+    @Test
+    fun `should return true when password is not valid`() {
```
AimanYosofiUlt:
```suggestion
    fun `should return false when password is not valid`() {
```

identity_domain/src/jvmTest/kotlin/net/thechance/mena/identity/domain/useCase/LoginUseCaseTest.kt
```diff
@@ -0,0 +1,124 @@
...
+        val result = loginUseCase.login(countryCode, number, password)
+
+        assertThat(result).isEqualTo(Unit)
+    }
+
+    @Test
+    fun `should throw UserIsBlockedException when user is blocked`() = runTest {
+        coEvery { authenticationRepository.isUserBlocked(any(), any()) } returns true
+        every { mobileNumberValidator.isValid(any(), any()) } returns true
+        coEvery { authenticationRepository.login(any(), any(), any()) } just Runs
```
AimanYosofiUlt: I think you can remove this line; it looks redundant

AimanYosofiUlt (commented):
Great work! Just please check the comments

identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/useCase/LoginUseCase.kt
```diff
@@ -0,0 +1,36 @@
...
+            number = number,
+            password = password
+        )
+    }
+
+    fun isMobileNumberValid(countryCode: String, number: String): Boolean {
+        return mobileNumberValidator.isValid(countryCode = countryCode, number = number)
+    }
+
+    fun isPasswordValid(password: String) = password.length >= 8
```
AimanYosofiUlt: I think you should store this value `8` inside a constant

identity_domain/src/jvmTest/kotlin/net/thechance/mena/identity/domain/useCase/LoginUseCaseTest.kt
poula22: which one?

identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/exception/Exception.kt
```diff
@@ -25,4 +25,4 @@ class InvalidCredentialsException(
     "user with mobile number: $countryCode$mobileNumber doesn't exist or password is incorrect"
 )
```
mohamedshemees, line 27: i don't think we have to have hardcoded messages here , maybe only throw it and react upon it in presentation with proper localized string resource , or we have this only for debugging ?

identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/exception/Exception.kt
```diff
@@ -25,4 +25,4 @@ class InvalidCredentialsException(
...
 
-class UserNeedsLoginException : AuthenticationException("user has no access to application")
```
mohamedshemees, line 28: i think SessionExpiredException would be better

identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/repository/AuthenticationRepository.kt
```diff
@@ -0,0 +1,8 @@
+package net.thechance.mena.identity.domain.repository
+
+interface AuthenticationRepository {
+    suspend fun login(countryCode: String, number: String, password: String)
+    suspend fun isLoggedIn(): Boolean
+    suspend fun isUserBlocked(countryCode: String, number: String): Boolean
```
mohamedshemees: i don't think we need this as being blocked will be determined server side , so i don't need a specific request , server can throw Rate limit exception on login end point

identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/repository/AuthenticationRepository.kt
```diff
@@ -0,0 +1,8 @@
+package net.thechance.mena.identity.domain.repository
+
+interface AuthenticationRepository {
+    suspend fun login(countryCode: String, number: String, password: String)
+    suspend fun isLoggedIn(): Boolean
```
mohamedshemees: is logged in , i don't think it has a specific use case as well , the flow will be like this  , user try to login , he logs in for first time , next time , checking if logged in or no (access token is expired or not ) will be checked in the process of navigating to home normally after calling getToken function, (and append it as a header)

identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/exception/Exception.kt
poula22: this log is only for debugging

identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/exception/Exception.kt
poula22: this line is removed it's now called unAuzorizedException.

identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/repository/AuthenticationRepository.kt
poula22: I will remove it

identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/repository/AuthenticationRepository.kt
poula22: I think we can handle it in local device also, so if user tried to login and we know that he is blocked, this request will not go to server.

mohamedshemees: I thought we said we can't trust device local time , it can be manipulated, or i have wrong info here ?

identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/useCase/validation/mobileNumber/MobileNumberValidator.kt
```diff
@@ -0,0 +1,44 @@
+package net.thechance.mena.identity.domain.useCase.validation.mobileNumber
+
+import net.thechance.mena.identity.domain.exception.InvalidCountryCodeException
+
+class MobileNumberValidator {
```
nabilBouzineDev, line 5: Great work @poula22, but I think this should be an interface rather than a concrete implementation.

poula22: Why should Mobile Validator have abstraction?

poula22: It's a helper class to login use case, so you treat it like utility class

nabilBouzineDev: That's true 

But it may be a scenario where we don't accept the leading digit for all mobile numbers or, we've added a third party library that does the validation. 

So, I think having an abstraction here will prevent us from breaking SOLID principles. 

what do you think?

nabilBouzineDev (approved):
Good work, @poula22 👍

ShroukMohamed16 (approved):

PR #79, opened 2025-09-17
Feature/IDN-16 Configure multi environment builds

eslam-99:
This PR introduces different values for different environment variables based on the selected Flavor/Scheme for both Andriod & IOS

Needed Setup:
 - in `local.properties` add:
 ```
BASE_URL_DEVELOPMENT=
BASE_URL_PRODUCTION=
BASE_URL_STAGING=
```

### DIFF
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -1,4 +1,5 @@
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+import java.util.Properties
 import java.io.ByteArrayOutputStream
 
 plugins {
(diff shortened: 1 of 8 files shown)


ThoriaHamdy99 (approved):

ABDULLAHHG (approved):

composeApp/build.gradle.kts
AhmadKharfan, line 1: Good job but consider moving the hardcoded strings flavor names, suffixes, BASE_URL keys, etc... into the version catalog (libs.versions.toml)

AhmadKharfan (approved):

composeApp/build.gradle.kts
eslam-99: Good point, will apply it

eslam-99: @AhmadKharfan 
After search, I found that version catalog not supporting `[strings]` section, only `[versions], [libraries], [bundles], [plugins], [metadata]`
So, it requires a plugin module as we implemented in previous season
what do you think?

AhmadKharfan: You're right the only solution is to put them in the version catalog which isn’t really ideal
As for the module I mentioned something like a build-logic module we won’t do it in this sprint we can skip it for now and plan to implement it in a future sprint

PR #81, opened 2025-09-17
Feature CHT 23 test for contacts repository

dev-bilal-azzam:
- create fake classes for `ContactsRepositoryImpl` such as `ContactsProvider` & `DataStore`
- test `ContactsRepositoryImpl`


core_chat_data/build.gradle.kts
```diff
@@ -24,6 +24,13 @@ kotlin {
             implementation(libs.androidx.datastore.preferences)
 
         }
+        commonTest.dependencies {
+            implementation(libs.kotlin.test)
+            implementation(kotlin("test-annotations-common"))
```
ZainabAlnajjar: you can add this to toml file for consistency

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImpl.kt
```diff
@@ -53,6 +55,7 @@ class ContactsRepositoryImpl(
             defaultException = { ContactSyncFailedException("Couldn't sync user contacts", it) }) {
             val contacts = getDeviceContacts()
             client.post(SYNC_CONTACTS_ENDPOINT) {
+                contentType(ContentType.Application.Json)
```
ZainabAlnajjar: why did you add this line? it is already specified in the HttpClientHolder file

core_chat_data/build.gradle.kts
dev-bilal-azzam: done

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImpl.kt
dev-bilal-azzam: Right, I removed it

core_chat_data/src/commonTest/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImplTest.kt
```diff
@@ -0,0 +1,408 @@
...
+    private val mockDataStore = FakeDataStore()
+    private val json = Json { ignoreUnknownKeys = true }
+    private val jsonHeaders = headersOf(
+        HttpHeaders.ContentType,
+        ContentType.Application.Json.toString()
+    )
+
+    private lateinit var httpClient: HttpClient
+    private lateinit var repository: ContactsRepositoryImpl
+    private val sampleContact = Contact(
```
ZainabAlnajjar: Since you are not depending on returning mocked data, but a fake data source, I think It would be better to extract these into fake data samples file and use the mappers to convert it to `ContactDto` that the api return, so the data won`t be hardcoded, also to make it easier to change the tests samples later without failing the tests

core_chat_data/src/commonTest/kotlin/net/thechance/mena/core_chat/data/contacts/utils.kt
```diff
@@ -0,0 +1,121 @@
...
+    )
+}
+
+fun MockRequestHandleScope.defaultContactsResponse(json: Json, headers: Headers) = respond(
+    content = successResponse<PagedDataDto<ContactDto>>(
+        json = json,
+        body = PagedDataDto(
+            data = listOf(
+                ContactDto(
+                    name = "Bilal Azzam",
```
ZainabAlnajjar: In the case I mentioned in previous comment, this should also be extracted

core_chat_data/src/commonTest/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImplTest.kt
```diff
@@ -0,0 +1,408 @@
...
+    fun setUp() {
+        mockContactsProvider.reset()
+        mockDataStore.reset()
+        httpClient = createHttpClient(jsonHeaders, json)
+
+        repository = createRepository(mockContactsProvider, mockDataStore, json, jsonHeaders)
+    }
+
+    @Test
+    fun `should return paged contacts when getUserContacts is called successfully`() = runTest {
```
ZainabAlnajjar, line 46: I recommend to follow the suggested pattern for test naming from previous seasons at the chance, and from clean code book: 
`functionName should return/do... when... `

So first thing we see where the error happened, also I believe the other features would also use it

core_chat_data/src/commonTest/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImplTest.kt
```diff
@@ -0,0 +1,408 @@
...
+    fun `should throw UnAuthorizedException when getUserContacts receives unauthorized response`() =
+        runTest {
+
+            repository = createRepository(
+                contactsProvider = mockContactsProvider,
+                contactsDataStore = mockDataStore,
+                json = json,
+                jsonHeaders = jsonHeaders,
+                contactsResponse = {
+                    respond(
```
ZainabAlnajjar: This response is large and repeated with few changes, i suggest to make fake response builder with default values, and change only the values that matter for the test

core_chat_data/src/commonTest/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImplTest.kt
ZainabAlnajjar: With a one empty line between the (Given, When and Then)

core_chat_data/src/commonTest/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImplTest.kt
```diff
@@ -0,0 +1,408 @@
...
+                    )
+                }
+            )
+
+
+            val exception = assertFailsWith<UnAuthorizedException> {
+
+                repository.getUserContacts(pageNumber = 1, pageSize = 10)
+            }
+            assertThat(exception.message).isEqualTo("Unauthorized")
```
ZainabAlnajjar: I think we don't need this assert, the thing that matters here is throwing the exception

Also if you needed to assert equal for strings, it is better to extract variable instead on hardcoding it twice, so the test won't fail from a typo

core_chat_data/src/commonTest/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImplTest.kt
```diff
@@ -0,0 +1,408 @@
...
+                )
+            }
+        )
+
+
+        val exception = assertFailsWith<UnknownException> {
+
+            repository.getUserContacts(pageNumber = 1, pageSize = 10)
+        }
+        assertThat(exception.message).isEqualTo("Server error")
```
ZainabAlnajjar: Same here

core_chat_data/src/commonTest/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImplTest.kt
```diff
@@ -0,0 +1,408 @@
...
+                    throw IOException("Network error")
+                }
+            )
+
+
+            val exception = assertFailsWith<ContactsFetchFailedException> {
+
+                repository.getUserContacts(pageNumber = 1, pageSize = 10)
+            }
+            assertThat(exception.message).isEqualTo("Couldn't get user contacts")
```
ZainabAlnajjar: This case for example, the test would fail if we changed the message, while it is not that important how it was written

It would be better to remove this part from other tests as well

gradle/libs.versions.toml
```diff
@@ -63,6 +72,9 @@ koin-annotations = { module = "io.insert-koin:koin-annotations", version.ref = "
 koin-ksp-compiler = { module = "io.insert-koin:koin-ksp-compiler", version.ref = "koin-annotations" }
 
 # permissions
+ktor-client-mock = { module = "io.ktor:ktor-client-mock", version.ref = "ktor" }
+mockk = { module = "io.mockk:mockk", version.ref = "mockk" }
+mockk-common = { module = "io.mockk:mockk-common", version.ref = "mockk" }
```
Joseph-Sameh-0: I saw that you didn't use it so why you don't remove it ?

core_chat_data/src/commonTest/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImplTest.kt
```diff
@@ -0,0 +1,408 @@
...
+                            pageSize = 10,
+                            totalItems = 0,
+                            totalPages = 1
+                        )
+                    ),
+                    status = HttpStatusCode.OK,
+                    headers = jsonHeaders
+                )
+            }
+        )
```
Joseph-Sameh-0: you can extract this part in a separate function at the end of the test class

Joseph-Sameh-0: you can give it default values and change what you want when you use it

Joseph-Sameh-0 (approved):

core_chat_data/src/commonTest/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImplTest.kt
```diff
@@ -405,4 +329,70 @@ class ContactsRepositoryImplTest {
             assertThat(mockDataStore.getUserSyncedState()).isFalse()
         }
 
+    @Test
+    fun `should return 10 contacts on first page`() = runTest {
```
ZainabAlnajjar: the test name is not clear, why it should return 10? maybe we changed the page size, what happens then? also we need to specify the function name, this way we can debug better if the test failed

core_chat_data/src/commonTest/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImplTest.kt
```diff
@@ -405,4 +329,70 @@ class ContactsRepositoryImplTest {
...
+            }
+        )
+
+        val result = repository.getUserContacts(pageNumber = 1, pageSize = 10)
+
+        assertThat(result.data).isEqualTo(page1Contacts.data?.map { it.toDomain() })
+    }
+
+    @Test
+    fun `should return 5 contacts on second page`() = runTest {
```
ZainabAlnajjar: same here the test is not clear 
always remember that the test is meant for debugging in future, and probably other developers will encounter this message is terminal, so it need to be clear with all needed details to pass it

core_chat_data/src/commonTest/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImplTest.kt
```diff
@@ -405,4 +329,70 @@ class ContactsRepositoryImplTest {
...
+            }
+        )
+
+        val result = repository.getUserContacts(pageNumber = 2, pageSize = 10)
+
+        assertThat(result.data).isEqualTo(page2Contacts.data?.map { it.toDomain() })
+    }
+
+    @Test
+    fun `should return empty list when requesting page beyond total pages`() = runTest {
```
ZainabAlnajjar: need to specify function name

core_chat_data/src/commonTest/kotlin/net/thechance/mena/core_chat/data/contacts/ContactsRepositoryImplTest.kt
Noor-Serry: I agree with @ZainabAlnajjar we should write the name of the function if there are more than one public function inside the class 
so I can know the failed function direct

core_chat_data/src/commonTest/kotlin/net/thechance/mena/core_chat/data/contacts/fakes/FakeContactsProvider.kt
```diff
@@ -0,0 +1,31 @@
+package net.thechance.mena.core_chat.data.contacts.fakes
+
+import com.bilalazzam.contacts_provider.ContactField
+import com.bilalazzam.contacts_provider.ContactsPermissionDeniedException
+import com.bilalazzam.contacts_provider.ContactsProvider
+import com.bilalazzam.contacts_provider.Contact as DeviceContact
+
+
+class FakeContactsProvider(
```
Noor-Serry, line 9: I think ContactsProvider can be mocked with mokkery https://mokkery.dev/

omer1998 (approved):
good job bilal, it seems good to me, just some functions need to be better in terms of naming

ZainabAlnajjar (approved):

PR #88, opened 2025-09-18
Feature/wlt 11 setup repo di and exceptions

hagarabobakr:
Added Koin DI setup using KSP for automatic injection and implemented safe API call handling with custom exceptions for better error management.


Ahmed7osny1 (approved):

wallet_data/src/commonMain/kotlin/net/thechance/mena/wallet/data/exceptions/EmptyResponse.kt
```diff
@@ -0,0 +1,6 @@
+package net.thechance.mena.wallet.data.exceptions
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+class EmptyResponse
```
Farah315, line 6: i think we can rename it a WalletResponse or something because it will be more clear when we replace the real response here

Al-Qassim (approved):

PR #102, opened 2025-09-18
Feature/WLT-6 Wallet screen UI fixes

MuhammedEdrees:
### PR Content:
- Fixed minor bugs and inconsistencies and bugs in wallet screen.

### DIFF
--- a/wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/utils/TextUtils.kt
+++ b/wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/utils/TextUtils.kt
@@ -1,0 +1,20 @@
+package net.thechance.mena.wallet.presentation.utils
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.text.TextStyle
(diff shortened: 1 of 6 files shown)


wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/utils/TextUtils.kt
```diff
@@ -0,0 +1,20 @@
...
+    style: TextStyle,
+): Dp {
+    val density = LocalDensity.current
+
+    return with(density) {
+        val lineHeightSp = style.lineHeight.takeOrElse { style.fontSize }
+        val lineHeightPx = lineHeightSp.toPx()
+        lineHeightPx.toDp()
+    }
+}
```
Al-Qassim: it will be better to remove the composable annotation and pass the density from outside as a parameter
no need to have the whole TextStyle as input, we only need the lineHeight in sp.
also, you can make the function simpler, it's whole job is to convert Sp to Dp, so you can make it an extenstion function on Sp

```
fun Sp.toDp(density: Density) = with(density) { this.toPx().toDp() }
```

Al-Qassim (approved):

Malak187 (approved):

PR #104, opened 2025-09-18
Feature/TRN-24 exception handling

iZyadNasser:
### This PR introduces:

- Currently needed domain exception (only no internet exception)
- Implementation of the safe api call function to handle network exceptions and error statuses (we currently map only the io exception to no internet exception, as for the rest of the exceptions and status codes we only log them for development purposes, **_also not that the currently logged error messages are temporary until the base network error model is created for backend and we can log the api message_**
- I also added a very simple custom logger for convenience, it now only has the error logger but we can add to it what we need as we continue developement


trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/util/safeApiCall.kt
```diff
@@ -1,17 +1,98 @@
...
+
+        in 400..499 -> {
+            when (result.status) {
+                HttpStatusCode.Unauthorized -> {
+                    logError(
+                        HANDLE_ERROR_STATUS_TAG,
+                        "Unauthorized",
+                        "Not authorized to do this action"
+                    )
+                    throw Exception()
```
ahmednasserzaza, line 45: Why not using our customized exceptions?

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/util/safeApiCall.kt
```diff
@@ -1,17 +1,98 @@
...
+        throw NoInternetException()
+    } catch (exception: UnresolvedAddressException) {
+        logError(SAFE_API_CALL_TAG, "UnresolvedAddressException", exception.message.toString())
+        throw NoInternetException()
+    } catch (exception: Exception) {
+        logError(SAFE_API_CALL_TAG, "Unknown exception", exception.message.toString())
+        throw exception
     }
 
     return handleErrorStatus(result)
```
ahmednasserzaza: I think this function should be renamed, because it's not handling the error status only, it handles the success status too

ahmednasserzaza (commented):
Good work 👏
Just check the comments please

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/util/safeApiCall.kt
iZyadNasser: We agreed that for now we only need to handle the no internet exception for the user to check their internet connection, any other exception or error status we will just give the user a generic message like "something went wrong", only the developer will know what the error was specifically using logs, we will add more custom exceptions as we need them in the future

ahmeddsamii (approved):

PR #112, opened 2025-09-19
Feature/wlt 12 real implementation for repo

hagarabobakr:
Implemented the real repository logic and connected it to the backend.

### DIFF
--- a/composeApp/src/androidMain/AndroidManifest.xml
+++ b/composeApp/src/androidMain/AndroidManifest.xml
@@ -1,7 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
+    <uses-permission android:name="android.permission.INTERNET"/>
+
     <application
+        android:usesCleartextTraffic="true"
         android:allowBackup="true"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
--- a/wallet_data/src/commonMain/kotlin/net/thechance/mena/wallet/data/utils/NetworkClient.kt
+++ b/wallet_data/src/commonMain/kotlin/net/thechance/mena/wallet/data/utils/NetworkClient.kt
@@ -78,7 +78,7 @@ class NetworkClient {
     }
 
     companion object {
-        private const val BASE_URL = "https://" //TODO: add base url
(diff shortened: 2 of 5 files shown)


composeApp/src/androidMain/AndroidManifest.xml
```diff
@@ -1,7 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
+    <uses-permission android:name="android.permission.INTERNET"/>
+
     <application
+        android:usesCleartextTraffic="true"
```
Al-Qassim, line 7: this line just allow the test to be made locally.
it essentially just let me connect to local host.
so no need for it here.

wallet_data/src/commonMain/kotlin/net/thechance/mena/wallet/data/utils/NetworkClient.kt
```diff
@@ -78,7 +92,9 @@ class NetworkClient {
     }
 
     companion object {
-        private const val BASE_URL = "https://" //TODO: add base url
+        private const val BASE_URL = "https://mena-dev.the-chance.net/"
+        private const val LOCAL_BASE_URL = "http://10.0.2.2:8080/"
```
Al-Qassim, line 96: the `BASE_URL` and `LOCAL_BASE_URL` should be moved to a file like local.properties or environment variables, not hard coded

and you should not use `LOCAL_BASE_URL`, only one variable should be used, which is `BASE_URL`, so we will not need to change the source code

wallet_data/src/commonMain/kotlin/net/thechance/mena/wallet/data/utils/NetworkClient.kt
```diff
@@ -78,7 +92,9 @@ class NetworkClient {
...
+        val ACCESS_TOKEN = "eyJhbGciOiJIUzI1NiJ9.eyJzdWIiOiIzOGMxNWY2MS1kNW"
```
Al-Qassim, line 97: this can stay here temporarily, until we receive the `getToken` function from the identity team

Al-Qassim (approved):

gehadepada (approved):

PR #122, opened 2025-09-19
Feature/IDN-7 Configure CI/CD

ABDULLAHHG:
### Slack Notifications
1 - build error (Android or IOS)
2 - Test coverage checks for only changed modules
### Build Flavors Added
1 - develop
2 - staging
### Deployment for Each Build Flavor
1 - on slack
2 - on firebase


AhmadKharfan (approved):

.github/workflows/ios-build.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: chmod +x gradlew
+
+      - name: Build KMP modules
+        run: ./gradlew $(for m in $(echo '${{ needs.detect.outputs.ios_modules }}' | jq -r '.[]'); do echo ":$m:linkDebugFrameworkIosSimulatorArm64"; done) --parallel --build-cache
+
+      - name: Build iOS app
+        if: needs.detect.outputs.has_ios_changes == 'true'
+        run: |
+          xcodebuild \
+            -project iosApp/iosApp.xcodeproj \
```
YoussefmSaber, line 40: you going to need to change the `iosApp.xcodeproj` to `iosApp.xcworkspace` so it's going to build the project using `cocoapods` libraries.

YoussefmSaber (approved):

.github/workflows/ios-build.yml
AhmadKharfan: Can’t switch to iosApp.xcworkspace yet because the project doesn’t have a Podfile or CocoaPods setup we can add it in a future sprint and will update the workflow then

.github/actions/build-release-apk/action.yml
```diff
@@ -0,0 +1,99 @@
...
+    - name: Make gradlew executable
+      run: chmod +x ./gradlew
+      shell: bash
+
+    - name: Version Bump
+      uses: chkfung/android-version-actions@v1.2.3
+      with:
+        gradlePath: composeApp/build.gradle.kts
+        versionCode: ${{ github.run_number }}
+        versionName: 1.0.${{ github.run_number }}
```
mostafa-n3ma: I don’t think it’s a good idea to hardcode 1 in the version name (1.0.${{ github.run_number }}). This makes the major and minor versions fixed, which could cause issues later when we need to bump them.

A more flexible approach would be to pull major and minor from gradle.properties (or another config file) and only use github.run_number for the patch/build number. For example:
`# gradle.properties
VERSION_MAJOR=1
VERSION_MINOR=0`

then you can do :
`    versionName: ${{ env.VERSION_MAJOR }}.${{ env.VERSION_MINOR }}.${{ github.run_number }}`


checkout this documentation :
[[Semantic Versioning](https://semver.org/?utm_source=chatgpt.com)]

.github/actions/build-release-apk/action.yml
```diff
@@ -0,0 +1,99 @@
...
+        fi
+        
+        APK_NAME="${VERSION_NAME}-v1.0.${{ github.run_number }}"
+        
+        mv composeApp/build/outputs/apk/${{ steps.convert.outputs.build_type_lower }}/release/composeApp-${{ steps.convert.outputs.build_type_lower }}-release-unsigned-signed.apk \
+           composeApp/build/outputs/apk/${{ steps.convert.outputs.build_type_lower }}/release/$APK_NAME.apk
+        
+        echo "APK_NAME=$APK_NAME" >> $GITHUB_ENV
+        echo "SIGNED_APK_PATH=composeApp/build/outputs/apk/${{ steps.convert.outputs.build_type_lower }}/release/$APK_NAME.apk" >> $GITHUB_ENV
+      shell: bash
```
mostafa-n3ma: consider enabling debugging here

mostafa-n3ma (approved):

ahmeddsamii (approved):

.github/workflows/test_coverage.yml
```diff
@@ -0,0 +1,89 @@
...
+          name: html-report-${{ matrix.module }}
+          path: ${{ matrix.module }}/build/reports/kover/html/
+
+      - uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: xml-report-${{ matrix.module }}
+          path: ${{ matrix.module }}/build/reports/kover/report.xml
+
+  codecov:
```
ZainabAlnajjar, line 52: Since we are using codecov, I suggest adding `codecov.yml` to specify what it does exactly, because it does not follow kover rules
We used codecov before and it was setting random test coverage percentage to each PR, therefor failing the PR although kover test coverage passed, the yml file solved it

AhmadKharfan: We’re only using Codecov to display the coverage report on the PR not to enforce thresholds the actual verification is already handled by Kover/KoverVerify which fails the build if the required percentage isn’t met also our coverage checks run per module rather than at the whole project level because of that we don’t really need a codecov.yml Codecov here is just for visualization

ZainabAlnajjar (approved):

Al-Qassim (approved):

AzizaHelmy (approved):

iZyadNasser (approved):

abdallahFc (approved):

Asiasama710 (approved):

FaresM0hamed (approved):

iBareq (approved):

PR #123, opened 2025-09-19
Feature/dkn 11 dukan unique name and categories

AsmaaKhaledGomaa:
Add a view model unit test for dukan unique name and categories


dukan_presentation/src/commonTest/kotlin/net/thechance/mena/dukan/presentation/viewModel/createDukan/CreateDukanViewModelTest.kt
```diff
@@ -357,38 +348,441 @@ class CreateDukanViewModelTest {
...
+    fun `onNameChanged SHOULD update next button state`() = runTest {
+        // Given
+        val name = fakeTestName()
+
+        // When
+        createDukanViewModel.onNameChanged(name)
+
+        // Then
+        assertTrue(true)
+    }
```
haideraljawasim, line 411: i don't get this test ? assert true true

ahmedNaser7 (approved):

PR #140, opened 2025-09-21
Feature/WLT-2 view my balance

Malak187:
View my balance story branch.


abdallahFc (approved):

wallet_data/build.gradle.kts
```diff
@@ -1,20 +1,77 @@
...
+            jvmTarget.set(JvmTarget.JVM_11)
+        }
+    }
     iosArm64()
     iosSimulatorArm64()
 
     sourceSets {
+        androidMain.dependencies {
+            //ktor
+            implementation(libs.ktor.client.android)
```
Al-Qassim: we can remove 
`implementation(libs.ktor.client.android)`

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/base/BaseViewModel.kt
```diff
@@ -0,0 +1,55 @@
...
+    protected fun sendEffect(effect: EFFECT) {
+        viewModelScope.launch {
+            _uiEffect.emit(effect)
+        }
+    }
+
+    protected fun <T> tryToExecute(
+        callee: suspend () -> T,
+        onSuccess: (suspend (T) -> Unit)? = null,
+        onError: (suspend (Throwable) -> Unit)? = null,
```
Al-Qassim: better to remove default value of null for `onError` and `onSuccess`

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/base/SnackBarState.kt
```diff
@@ -0,0 +1,10 @@
+package net.thechance.mena.wallet.presentation.base
+
+import org.jetbrains.compose.resources.StringResource
+
+data class SnackBarState(
+    val isVisible: Boolean = false,
+    val titleRes: StringResource? = null,
+    val messageRes: StringResource? = null,
+    val isSuccess: Boolean = true
```
Al-Qassim, line 9: it's better to have `iconRes` instead of `isSuccess`, since we may use  icons other than tick and cross in the snack bar

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/base/UiState.kt
```diff
@@ -0,0 +1,12 @@
+package net.thechance.mena.wallet.presentation.base
+
+sealed class UiState<out T> {
+    object Idle : UiState<Nothing>()
+    object Loading : UiState<Nothing>()
+    data class Success<T>(val data: T) : UiState<T>()
+    data class Error(val throwable: Throwable? = null) : UiState<Nothing>()
```
Al-Qassim, line 7: why `throwable` is nullable?, it should not be null.
also, a uiState should not hold the error object, it should hold a custom ui error state that will be mapped to an appropriate user-friendly error message

Al-Qassim (approved):

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/base/SnackBarState.kt
MuhammedEdrees: Hmm, I think it's fine to leave it for now since there is no other snackbar icons in the figma design and we can change it later if needed.

EL-MANCY (approved):

Ahmed7osny1 (approved):

PR #149, opened 2025-09-22
Feature/FTH-33-sur-view-model-unit-test

marwanqashwa:
**Testing :**

* Added the `mokkery` plugin and relevant libraries (`turbine`, `mokkery.core`, `kotlinx.coroutines.test`) to the build configuration in `faith_presentation/build.gradle.kts` 

* Added a comprehensive test suite for `SurViewModel` in `SurViewModelTest.kt`, covering initialization, success/error states, and navigation effects, using `mokkery` for mocking and `turbine` for effect testing.

**SurViewModel improvements :**

* Modified the `SurViewModel` constructor to accept a `CoroutineDispatcher` (defaulting to `Dispatchers.IO`) for easier testing and better control of coroutine context.

### DIFF
--- a/faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
+++ b/faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
@@ -1,0 +1,186 @@
+package net.thechance.mena.faith.presentation.feature.quran.sur
+
+import app.cash.turbine.test
+import dev.mokkery.MockMode
+import dev.mokkery.answering.returns
+import dev.mokkery.answering.throws
+import dev.mokkery.everySuspend
+import dev.mokkery.mock
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.StandardTestDispatcher
+import kotlinx.coroutines.test.resetMain
+import kotlinx.coroutines.test.runTest
+import kotlinx.coroutines.test.setMain
+import mena.faith_presentation.generated.resources.Res
+import mena.faith_presentation.generated.resources.ic_al_baqarah
+import mena.faith_presentation.generated.resources.ic_al_fatihah
+import net.thechance.mena.faith.domain.entity.Surah
+import net.thechance.mena.faith.domain.repository.QuranRepository
+import org.junit.Test
+import kotlin.test.AfterTest
+import kotlin.test.BeforeTest
+import kotlin.test.assertEquals
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class SurViewModelTest {
+
+    private lateinit var viewModel: SurViewModel
+    private lateinit var quranRepository: QuranRepository
+    private val testDispatcher = StandardTestDispatcher()
+
+    @BeforeTest
+    fun setUp() {
+        Dispatchers.setMain(testDispatcher)
+        quranRepository = mock(mode = MockMode.autofill)
+        everySuspend { quranRepository.getAllSur() } returns surList
+        viewModel = SurViewModel(quranRepository, testDispatcher)
+    }
+
+    @AfterTest
+    fun tearDown() {
+        Dispatchers.resetMain()
+    }
+
+    @Test
+    fun `initialize view model should set empty state when getAllSur returns no data`() = runTest {
+        // Given
+        everySuspend { quranRepository.getAllSur() } returns emptyList()
+
+        // When
+        viewModel = SurViewModel(quranRepository, testDispatcher)
+        testDispatcher.scheduler.advanceUntilIdle()
+
+        // Then
+        assertEquals(emptyUiState, viewModel.uiState.value)
+    }
+
+    @Test
+    fun `initialize view model should set success state when getAllSur returns data`() =
+        runTest {
+            // Given
+            everySuspend { quranRepository.getAllSur() } returns surList
+
+            // When
+            viewModel = SurViewModel(quranRepository, testDispatcher)
+            testDispatcher.scheduler.advanceUntilIdle()
+
+            // Then
+            assertEquals(
+                emptyUiState.copy(sur = surahUiList, errorMessage = null, isLoading = false),
+                viewModel.uiState.value
+            )
+        }
+
+    @Test
+    fun `initialize view model should handle error state when getAllSur fails`() = runTest {
+        // Given
+        everySuspend { quranRepository.getAllSur() } throws Exception("Error")
+
+        // When
+        viewModel = SurViewModel(quranRepository, testDispatcher)
+        testDispatcher.scheduler.advanceUntilIdle()
+
+        // Then
+        assertEquals(
+            emptyUiState.copy(errorMessage = "Error", isLoading = false),
+            viewModel.uiState.value
+        )
+    }
+
+    @Test
+    fun `onSurahClick should navigate to surah details screen with surah id and name`() = runTest {
+        // Given
+        testDispatcher.scheduler.advanceUntilIdle()
+
+        // When
+        viewModel.uiEffect.test {
+            viewModel.onSurahClick(AL_FATIHAH_ID, AL_FATIHAH_NAME)
+
+            // Then
+            assertEquals(
+                SurEffect.NavigateToSurahDetails(AL_FATIHAH_ID, AL_FATIHAH_NAME),
+                awaitItem()
+            )
+        }
+    }
+
+    @Test
+    fun `onBackClick should navigate back`() = runTest {
+        // Given
+        testDispatcher.scheduler.advanceUntilIdle()
+
+        // When
+        viewModel.uiEffect.test {
+            viewModel.onBackClick()
+
+            // Then
+            assertEquals(
+                SurEffect.NavigateBack,
+                awaitItem()
+            )
+        }
+    }
+
+    @Test
+    fun `onBookmarkClick should navigate to bookmark screen`() = runTest {
+        // Given
+        testDispatcher.scheduler.advanceUntilIdle()
+
+        // When
+        viewModel.uiEffect.test {
+            viewModel.onBookmarkClick()
+
+            // Then
+            assertEquals(
+                SurEffect.NavigateToBookmark,
+                awaitItem()
+            )
+        }
+    }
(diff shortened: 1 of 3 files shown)


faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
```diff
@@ -0,0 +1,186 @@
...
+
+    private lateinit var viewModel: SurViewModel
+    private lateinit var quranRepository: QuranRepository
+    private val testDispatcher = StandardTestDispatcher()
+
+    @BeforeTest
+    fun setUp() {
+        Dispatchers.setMain(testDispatcher)
+        quranRepository = mock(mode = MockMode.autofill)
+        everySuspend { quranRepository.getAllSur() } returns surList
```
FaresM0hamed: I think this stub here is useless because in each test function you re-stub this function with different return data , So i can't see what is the benefit for this stub , do you agree with me ?

FaresM0hamed: Do we need this beforeTest here?
I think we can initialize viewmodel and make repo mocking one time only , we don't need to make this before each test function  , So i recommend to remove this `setup` function , and initiate repo and viewmodel outside in normal way , what do you think ?

marwanqashwa: Yes, I agree with you. They were there for a purpose that no longer exists.

marwanqashwa: Both of them no longer make sense, since the dispatcher is now being passed to the ViewModel from the outside. Also, I stub getAllSur and create a new instance of the ViewModel whenever I need it, except for the effect, which doesn’t need a fresh one.

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
```diff
@@ -0,0 +1,173 @@
+package net.thechance.mena.faith.presentation.feature.quran.sur
+
+import app.cash.turbine.test
+import dev.mokkery.MockMode
+import dev.mokkery.answering.returns
+import dev.mokkery.answering.throws
+import dev.mokkery.everySuspend
+import dev.mokkery.mock
+import kotlinx.coroutines.Dispatchers
```
yasserahmed10: It's better to remove unused import

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
```diff
@@ -0,0 +1,173 @@
...
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.StandardTestDispatcher
+import kotlinx.coroutines.test.resetMain
```
yasserahmed10: It's better to remove unused import

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
```diff
@@ -0,0 +1,173 @@
...
+import kotlinx.coroutines.test.runTest
+import kotlinx.coroutines.test.setMain
```
yasserahmed10: It's better to remove unused import

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
```diff
@@ -0,0 +1,173 @@
...
+import mena.faith_presentation.generated.resources.Res
+import mena.faith_presentation.generated.resources.ic_al_baqarah
+import mena.faith_presentation.generated.resources.ic_al_fatihah
+import net.thechance.mena.faith.domain.entity.Surah
+import net.thechance.mena.faith.domain.repository.QuranRepository
+import org.junit.Test
+import kotlin.test.AfterTest
+import kotlin.test.BeforeTest
```
yasserahmed10: It's better to remove unused import

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
```diff
@@ -0,0 +1,173 @@
...
+import kotlin.test.assertEquals
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class SurViewModelTest {
+
+    private val quranRepository: QuranRepository = mock(MockMode.autofill)
+    private val testDispatcher = StandardTestDispatcher()
+    private var viewModel = SurViewModel(quranRepository, testDispatcher)
```
yasserahmed10, line 25: You can replace it with `private lateinit var viewModel: SurViewModel` because you  Initialized the ViewModel above

yasserahmed10 (approved):
Great Work Marwan 👏

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
```diff
@@ -0,0 +1,173 @@
...
+
+    @Test
+    fun `initialize view model should set empty state when getAllSur returns no data`() = runTest {
+        // Given
+        everySuspend { quranRepository.getAllSur() } returns emptyList()
```
fatmahgazy, line 30: You don’t need to use `everySuspend { quranRepository.getAllSur() } returns emptyList()` 
you can run the tests without it and they won’t fail, Mokkery handles something like this

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
```diff
@@ -0,0 +1,173 @@
...
+
+            // Then
+            assertEquals(
+                emptyUiState.copy(sur = surahUiList, errorMessage = null, isLoading = false),
+                viewModel.uiState.value
+            )
+        }
+
+    @Test
+    fun `initialize view model should handle error state when getAllSur fails`() = runTest {
```
fatmahgazy, line 58: I think we agreed not to test errors for now since they might be removed later, and we'd have to update the tests again after handling exceptions. What do you think?

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
marwanqashwa: I added it so I don’t have to create a new instance in every method, like when testing the effect. What do you think?

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
marwanqashwa: Yes, I know, but I kept it to make this test clearer. If I remove it, reading the test would be less obvious. What do you think?

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
marwanqashwa: Yes, but for now the error still exists in this form in the ViewModel, so I need to test it. Once it’s fixed, the change will be simple. I don’t think it’s an issue. What do you think?

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
```diff
@@ -0,0 +1,168 @@
...
+import net.thechance.mena.faith.domain.repository.QuranRepository
+import org.junit.Test
+import kotlin.test.assertEquals
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class SurViewModelTest {
+
+    private val quranRepository: QuranRepository = mock(MockMode.autofill)
+    private val testDispatcher = StandardTestDispatcher()
+    private var viewModel = SurViewModel(quranRepository, testDispatcher)
```
mohannadahmed00, line 25: It might be better to move the viewModel setup into a @ BeforeEach method.
That way every test starts with a fresh SurViewModel, so you don’t have to manually create it each time.

mohannadahmed00 (changes requested):

yasserahmed10 (approved):

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
marwanqashwa: I don’t think I need this instance, because I don’t always need a new one. I only need it in some tests, not all of them. Since I stub different repository behavior each time, I create a new instance in the methods where it’s required. What do you think?

BassantM (approved):

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
Mostafa-alsaygh: i agree with mohannad

mohannadahmed00 (approved):

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/sur/SurViewModelTest.kt
FaresM0hamed: i agree with marwan

FaresM0hamed (approved):

fatmahgazy (approved):

PR #150, opened 2025-09-24
test/FTH-34-surah-view-model

fatmahgazy:
**Unit Tests for the SurahViewModel** 
--------------------------------------------------------
 
- Added `SurahViewModelTest.kt` with comprehensive test cases covering various scenarios, including initialization, UI state updates, and user interactions.
- Updated `SurahViewModel.kt` to accept a `CoroutineDispatcher` for better testability.
- Modified `ClipboardManager` (common, Android, and iOS) to be an `open` class to allow for mocking in tests.
- Included necessary testing dependencies (`turbine`, `mokkery-core`, `kotlinx-coroutines-test`) in `build.gradle.kts`.


faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,381 @@
...
+            assertEquals(SurahScreenEffect.ShareAyah(EMPTY_STRING), effect)
+        }
+    }
+
+    companion object {
+        const val DEFAULT_SURAH_ID = 1
+        const val DEFAULT_SURAH_NAME = "Al-Fatiha"
+        const val TEST_SURAH_ID = 2
+        const val TEST_SURAH_NAME = "Al-Baqarah"
+        const val LARGE_SURAH_ID = 5
```
yasserahmed10: You can remove this const val as it is not used

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,381 @@
...
+import kotlin.test.assertNull
+import kotlin.test.assertTrue
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class SurahViewModelTest {
+
+    private val testDispatcher = StandardTestDispatcher()
+    private val quranRepository: QuranRepository = mock(mode = MockMode.autofill)
+    private val clipboardManager: ClipboardManager = mock(mode = MockMode.autofill)
+
```
yasserahmed10, line 22: You can make a setup method instead of repeating this  `everySuspend { quranRepository.getAyatOfSurah(any()) } returns emptyList()` in every test case
```
@BeforeTest
fun setup() {
    everySuspend { quranRepository.getAyatOfSurah(any()) } returns emptyList()
}

```

yasserahmed10 (approved):
Well Done Fatma 👏

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,381 @@
...
+            surahId = TEST_SURAH_ID,
+            surahName = DEFAULT_SURAH_NAME,
+            dispatcher = testDispatcher,
+            quranRepository = quranRepository,
+            clipboardManager = clipboardManager
+        )
+
+        // Then
+        assertEquals(TEST_SURAH_ID, testViewModel.uiState.value.surahId)
+    }
```
FaresM0hamed: I think this test case not give any value for this reasons:
1- you don't need surahId in uiState because it come from navigation and provided to viewmodel by DI , So actually you test DI here 

2- you don't use your stup here , I know if you remove it , It possible make test case fail , in Mockk we used relaxed , So i recommend to find some thing similar in mokkery .

yasserahmed10 (approved):

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
```
FaresM0hamed: I think this test case not give any value for this reasons:
1- you don't need surahId in uiState because it come from navigation and provided to viewmodel by DI , So actually you test DI here

2- you don't use your stup here , I know if you remove it , It possible make test case fail , in Mockk we used relaxed , So i recommend to find some thing similar in mokkery .

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
+    @Test
+    fun `SurahViewModel should return correct surah name when viewModel is created`() = runTest {
+        // When
+        val testViewModel = SurahViewModel(
+            surahId = DEFAULT_SURAH_ID,
+            surahName = TEST_SURAH_NAME,
+            dispatcher = testDispatcher,
+            quranRepository = quranRepository,
+            clipboardManager = clipboardManager
+        )
```
FaresM0hamed: I recommend to extract this outside test function to make it more readable

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
+            surahId = DEFAULT_SURAH_ID,
+            surahName = DEFAULT_SURAH_NAME,
+            dispatcher = testDispatcher,
+            quranRepository = quranRepository,
+            clipboardManager = clipboardManager
+        )
+
+        // Then
+        assertTrue(testViewModel.uiState.value.ayatOfSurah.isEmpty())
+    }
```
FaresM0hamed: In my opinion we could delegate testing on failure cases after we implement error states , Cause all of this will removed later

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
+
+    @Test
+    fun `onBackClick should emit NavigateBack effect when called`() = runTest {
+        // When & Then
+        viewModel.uiEffect.test {
+            viewModel.onBackClick()
+            val effect = awaitItem()
+            assertEquals(SurahScreenEffect.NavigateBack, effect)
+        }
+    }
```
FaresM0hamed: Let us make test function name in more English way like 
`onBackClick should navigate back when it called` 
your way not wrong ,At the end it's coding style and preferences

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
+
+    @Test
+    fun `onAyahLongPress should return true for isAyahActionButtonsVisible when called`() = runTest {
```
FaresM0hamed: Naming here make some confusion , I think if we name it some thing like 
`onAyahLongPress should show action buttons when it called`  it will be more clear , what do you think ?

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    @Test
+    fun `onAyahLongPress should return ayah index for selectedAyahIndex when called`() = runTest {
+        // When
+        viewModel.onAyahLongPress(TEST_AYAH_CONTENT, TEST_AYAH_INDEX)
+
+        // Then
+        assertEquals(TEST_AYAH_INDEX, viewModel.uiState.value.selectedAyahIndex)
+    }
```
FaresM0hamed, line 83: I think it should return ayahNumber not Index 
is Index == Number here ?

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
+
+    @Test
+    fun `onAyahLongPress should return negative value when called with negative index`() = runTest {
+        // When
+        viewModel.onAyahLongPress(TEST_AYAH_CONTENT, NEGATIVE_AYAH_INDEX)
+
+        // Then
+        assertEquals(NEGATIVE_AYAH_INDEX, viewModel.uiState.value.selectedAyahIndex)
+    }
```
FaresM0hamed, line 109: How could ayahNumber be in negative ?

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
+            // Given
+            viewModel.onAyahLongPress(TEST_AYAH_CONTENT, TEST_AYAH_INDEX)
+
+            // When
+            viewModel.onDismissActionButtons()
+
+            // Then
+            assertFalse(viewModel.uiState.value.isAyahActionButtonsVisible)
+        }
+
```
FaresM0hamed, line 124: same one in naming , its recommend to make it more describable like 
 onDismissActionButtons should hide action buttons when it called

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
+    fun `onDismissActionButtons should return false when ayah was long pressed`() = runTest {
+        // Given
+        viewModel.onAyahLongPress(TEST_AYAH_CONTENT, TEST_AYAH_INDEX)
+
+        // When
+        viewModel.onDismissActionButtons()
+
+        // Then
+        assertFalse(viewModel.uiState.value.isAyahActionButtonsVisible)
+    }
```
FaresM0hamed, line 151: There is some thing not logical here , It does't make sense onDismissActionButtons will be true when any ayah selected in any scenario , what do you think ?

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
+        // Given
+        viewModel.onAyahLongPress(TEST_AYAH_CONTENT, TEST_AYAH_INDEX)
+
+        // When
+        viewModel.onBookmarkClick(TEST_AYAH_NUMBER)
+
+        // Then
+        assertFalse(viewModel.uiState.value.isAyahActionButtonsVisible)
+    }
+
```
FaresM0hamed, line 152: same comment in naming

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
+    fun `onBookmarkClick should return null for selectedAyahIndex when called`() = runTest {
+        // Given
+        viewModel.onAyahLongPress(TEST_AYAH_CONTENT, TEST_AYAH_INDEX)
+
+        // When
+        viewModel.onBookmarkClick(TEST_AYAH_NUMBER)
+
+        // Then
+        assertNull(viewModel.uiState.value.selectedAyahIndex)
+    }
```
FaresM0hamed, line 178: Why return null here , actually i need surahId and ayahNumber

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
+
+    @Test
+    fun `onBookmarkClick should return false for isAyahActionButtonsVisible when called with zero ayah number`() = runTest {
```
FaresM0hamed: same comment in naming

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
+
+        // When
+        viewModel.onBookmarkClick(ZERO_AYAH_NUMBER)
+
+        // Then
+        assertFalse(viewModel.uiState.value.isAyahActionButtonsVisible)
+    }
+
+    @Test
+    fun `onShareClick should return false for isAyahActionButtonsVisible when called`() = runTest {
```
FaresM0hamed: same comment in naming here also

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
+    fun `onShareClick should return null for selectedAyahIndex when called`() = runTest {
+        // Given
+        viewModel.onAyahLongPress(TEST_AYAH_CONTENT, TEST_AYAH_INDEX)
+
+        // When
+        viewModel.onShareClick(TEST_AYAH_CONTENT)
+
+        // Then
+        assertNull(viewModel.uiState.value.selectedAyahIndex)
+    }
```
FaresM0hamed: If it will be null , so what  will be shared here ?

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
+    fun `onShareClick should emit ShareAyah effect with empty content when called with empty string`() = runTest {
+        // When & Then
+        viewModel.uiEffect.test {
+            viewModel.onShareClick(EMPTY_STRING)
+            val effect = awaitItem()
+            assertEquals(SurahScreenEffect.ShareAyah(EMPTY_STRING), effect)
+        }
+    }
+
+    companion object {
```
FaresM0hamed: Make it private cause you will not use it in any where unless this file

FaresM0hamed (changes requested):
Great Effort ya Fatma , Just small comments will make code better

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
marwanqashwa: The error still exists in the ViewModel in this form, so Once changed, the change will be simple. I don’t see an issue. What do you think?

FaresM0hamed: “We don’t yet know what error states will be added. When I reviewed this test case, I noticed it doesn’t include a value for the current time. I suggest that we only create such cases when we specifically need to test failure scenarios. Do you agree?

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,234 @@
...
+import kotlin.test.assertTrue
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class SurahViewModelTest {
+
+    private val testDispatcher = StandardTestDispatcher()
+    private val quranRepository: QuranRepository = mock(mode = MockMode.autofill)
+    private val clipboardManager: ClipboardManager = mock(mode = MockMode.autofill)
+
+    private fun createTestViewModel(
```
mohannadahmed00, line 23: It might be better to move the viewModel setup into a @ BeforeEach method.
That way every test starts with a fresh SurahViewModel, so you don’t have to manually create it each time, and you avoid any sneaky state carrying over between tests.

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
marwanqashwa: But this error currently exists in a way that, when an exception occurs, the error state is triggered with this message. Since that’s the case, the test should verify the correct behavior. Do you see what I mean?

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,234 @@
...
+
+    @Test
+    fun `onBackClick should navigate back when it called`() = runTest {
+        // Given
+        val testViewModel = createTestViewModel()
+
+        // When & Then
+        testViewModel.uiEffect.test {
+            testViewModel.onBackClick()
+            val effect = awaitItem()
```
mohannadahmed00, line 44: You don’t really need the `val effect = awaitItem()` line, just pass `awaitItem()` directly into `assertEquals` to keep the test concise.

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,234 @@
...
+
+        // When
+        testViewModel.onAyahLongPress(TEST_AYAH_CONTENT, TEST_AYAH_INDEX)
+
+        // Then
+        assertTrue(testViewModel.uiState.value.isAyahActionButtonsVisible)
+    }
+
+    @Test
+    fun `onAyahLongPress should return ayah content for selectedAyah when called`() = runTest {
```
mohannadahmed00, line 62: This test and the next one cover the same scenario. You could combine them into one test with two assertions, one for the index and one for the content to keep things more concise.

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,234 @@
...
+
+            // When
+            testViewModel.onDismissActionButtons()
+
+            // Then
+            assertFalse(testViewModel.uiState.value.isAyahActionButtonsVisible)
+        }
+
+    @Test
+    fun `onDismissActionButtons should return empty string for selectedAyah when called`() =
```
mohannadahmed00: It doesn’t return the value it updates the property to an empty string.

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,234 @@
...
+
+            // When
+            testViewModel.onDismissActionButtons()
+
+            // Then
+            assertEquals(EMPTY_STRING, testViewModel.uiState.value.selectedAyah)
+        }
+
+    @Test
+    fun `onDismissActionButtons should hide action buttons when called after ayah long press`() = runTest {
```
mohannadahmed00: This test overlaps with the `onDismissActionButtons should hide action buttons when it called` test, they’re both checking the same behavior.

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,234 @@
...
+        // When
+        testViewModel.onShareClick(TEST_AYAH_CONTENT)
+
+        // Then
+        assertFalse(testViewModel.uiState.value.isAyahActionButtonsVisible)
+    }
+
+
+    @Test
+    fun `onShareClick should return ayah content for selectedAyah when called`() = runTest {
```
mohannadahmed00: It doesn’t return the value; it updates the property.

faith_presentation/src/commonMain/kotlin/net/thechance/mena/faith/presentation/util/ClipBoardManager.kt
```diff
@@ -1,6 +1,6 @@
 package net.thechance.mena.faith.presentation.util
 
-expect class ClipboardManager {
-    fun copy(text: String)
+expect open class ClipboardManager {
```
mohannadahmed00, line 3: Is there a reason this class and function are marked `open`? It doesn’t look like we need inheritance or overriding here.

mohannadahmed00 (changes requested):

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
fatmahgazy: I see your point about using @BeforeEach, but in our case most tests need to pass different data into the ViewModel. Keeping the createTestViewModel() call inside each test gives us more flexibility and makes the setup explicit for each scenario

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
fatmahgazy: These are not the same one of them for content and index, but I prefer keeping them as separate tests. This way if one assertion fails, I’ll immediately know which part is broken. It makes the tests more clear and easier to debug What do you think?

faith_presentation/src/commonMain/kotlin/net/thechance/mena/faith/presentation/util/ClipBoardManager.kt
fatmahgazy: This class is `open`  to can be mocked in tests. Since I inject `ClipboardManager` into the ViewModel constructor, it needs to be mockable, and in Kotlin classes are final by default unless marked `open`.

mohannadahmed00 (approved):

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,220 @@
...
+        }
+    }
+
+    @Test
+    fun `onAyahLongPress should show action buttons when it called`() = runTest {
+        // Given
+        val testViewModel = createTestViewModel()
+
+        // When
+        testViewModel.onAyahLongPress(TEST_AYAH_CONTENT, TEST_AYAH_INDEX)
```
BassantM, line 55: Can u add named arguments for parameters to be clearer? What do u think

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,220 @@
...
+        assertTrue(testViewModel.uiState.value.isAyahActionButtonsVisible)
+    }
+
+    @Test
+    fun `onAyahLongPress should return ayah content for selectedAyah when called`() = runTest {
+        // Given
+        val testViewModel = createTestViewModel()
+
+        // When
+        testViewModel.onAyahLongPress(SELECTED_AYAH_CONTENT, TEST_AYAH_INDEX)
```
BassantM, line 67: Same, kindly use named argument

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
```diff
@@ -0,0 +1,220 @@
...
+        val testViewModel = createTestViewModel()
+
+        // When & Then
+        testViewModel.uiEffect.test {
+            testViewModel.onShareClick(AYAH_TO_SHARE)
+            assertEquals(SurahScreenEffect.ShareAyah(AYAH_TO_SHARE), awaitItem())
+        }
+    }
+
+    private companion object {
```
BassantM, line 205: u can make it private const companion object to reduce redundancy of constant

fatmahgazy: Modifier 'const' is not applicable to 'companion object'

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
fatmahgazy: It only contains fake data, so I think it’s already clear from the name that it’s for testing purposes 
so what do you think about leave it as it is?
if you prefer that , i have't any problem to change

FaresM0hamed: Agree with Fatma

faith_presentation/src/test/java/net/thechance/mena/faith/presentation/feature/quran/surah/SurahViewModelTest.kt
FaresM0hamed: Agree with Fatma

FaresM0hamed (approved):

BassantM (approved):
Great job

PR #158, opened 2025-09-25
Feature trn 30 add video description screen

HendSayed25:
# What's News ?

- Implement add video description screen 
- Implement screen view model , states ,effects and interaction listener
- Localize all Strings in Screen 
- Use Custom Save state handle rather than main class to can test it 
- Add Args constant object to contains all arguments will send to screens 
- Add Unit test for view model 

Extra point :  Update initialization of logger variable in base view model to prevent start koin in test file

# Short Video summarize what I said :  

https://github.com/user-attachments/assets/f6ed54a8-1468-45b8-a8c0-c2166ad9971b

### DIFF
--- a/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/video_description/VideoDescriptionScreen.kt
+++ b/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/video_description/VideoDescriptionScreen.kt
@@ -1,0 +1,203 @@
+package net.thechance.mena.trends.presentation.screen.video_description
+
+import androidx.compose.animation.animateColorAsState
+import androidx.compose.foundation.background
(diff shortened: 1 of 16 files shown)


trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/video_description/VideoDescriptionScreen.kt
```diff
@@ -0,0 +1,203 @@
...
+        state = state,
+        listener = viewModel,
+    )
+}
+
+@Composable
+private fun VideoDescriptionContent(
+    state: VideoDescriptionState,
+    listener: VideoDescriptionInteractionListener,
+    modifier: Modifier = Modifier,
```
fatimafaisalghazi, line 76: i dont think it is a good idea to pass modefier to a screen 
what do u think ?

PR #163, opened 2025-09-25
feature/DKN-56-refactor-the-navigation

ahmedNaser7:
# Refactor Navigation in Dukan Feature and Add Dukan API Interface

## Description
This PR refactors the navigation flow within the **Dukan** feature by introducing a new approach using the `DukanNavigator` interface.

### Changes:
- **Navigation refactor:**  
  All navigation logic is now handled through the `DukanNavigator` interface, allowing the ViewModel to control navigation in a clean and testable way.

- **Unit testing:**  
  Added comprehensive unit tests for navigation to ensure correct behavior across different ViewModel states.

- **API integration:**  
  Added `DukanApiInterface` and his implementation to represent the entry point of Dukan feature

- **Code improvements:**  
  Updated related classes and methods to work with the new navigation pattern and improve maintainability.

### Benefits:
- Enhances **testability** of navigation logic.
- Decouples **ViewModel** from UI.

### Possible Next steps :
- use the same way to display snackbar


dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/navigation/DukanNavigatorImpl.kt
```diff
@@ -0,0 +1,47 @@
...
+import kotlin.time.Clock
+import kotlin.time.ExperimentalTime
+
+
+class DukanNavigatorImpl(
+    override val startDestination: DukanRoute
+) : DukanNavigator {
+
+    private val _dukanEffects = MutableSharedFlow<DukanEffect>()
+    override val dukanEffects = _dukanEffects.asSharedFlow()
```
Ahmed-AlMancy, line 17: This is not used at all maybe we should reference it rather than _dukanEffects right?

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/viewModel/cropImage/ImageCropViewModel.kt
```diff
@@ -56,7 +66,11 @@ class ImageCropViewModel() : BaseViewModel<
     }
 
     private fun onCropImageSuccess(selectedImage: ImageBitmap) {
-        emitEffect(ImageCropEffects.NavigateBack(selectedImage))
+        viewModelScope.launch(
+            context = defaultDispatcher,
+        ) {
+            _effect.emit(ImageCropEffects.NavigateBack(selectedImage))
+        }
     }
```
haideraljawasim, line 74: hmm , so it will not replace the effects entirely  , and we still need to use them, for some one didn't write the code seeing the navigation here in different way will make no sense

ahmedNaser7: the probelm here in `ImageCropScreen` , when the screen observe on effect it used to call function `onImageCrop` not to use a real navigate to another screen , its big diffrence

Ahmed-AlMancy: i think that because we are passing selectedImage but in other screens its navigate(dukanRoute.X), so that is why he used effect.emit there, like you said it didn't eliminate the effects entirely or replace it

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/navigation/DukanNavigatorImpl.kt
ahmedNaser7: i cant understand your point , please can you explain more ?

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/viewModel/mainScreen/MainViewModel.kt
```diff
@@ -5,15 +5,19 @@ import kotlinx.coroutines.Dispatchers
...
+import net.thechance.mena.dukan.presentation.navigation.DukanNavigator
 import net.thechance.mena.dukan.presentation.viewModel.base.BaseViewModel
 import net.thechance.mena.dukan.presentation.viewModel.mainScreen.MainScreenUiState.DukanStatusUi
 
 class MainViewModel(
     private val dukanRepository: DukanRepository,
-    dispatcher: CoroutineDispatcher = Dispatchers.IO
-) : BaseViewModel<MainScreenUiState, MainEffect>(
+    dispatcher: CoroutineDispatcher = Dispatchers.IO,
+    navigator: DukanNavigator
```
YoussefmSaber: make the dispatcher the last constructor

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/viewModel/cropImage/ImageCropViewModel.kt
haideraljawasim: @ahmedNaser7 so , it will handle one way of navigation and later if we used the api to navigate to another feature for example also we will need to use the effects , I prefer we  keeping the effects  and the old way of navigation  because it gives more flexibility

ahmedNaser7: first point , for the example you talking about , here we use effect for just calling a `onImageCrop` , why we use effect when we can handl it using state of  `ImageCropViewModel ` its all about using flow 
``` @Composable
fun ImageCropScreen(
    selectedImage: ImageSrc?,
    onImageCrop: (ImageBitmap) -> Unit,
    viewModel: ImageCropViewModel = koinViewModel()
) {
    val state = viewModel.state.collectAsStateWithLifecycle()

    ObserveAsEffect(viewModel.effect) { effect ->
        when (effect) {
            is ImageCropEffects.NavigateBack -> onImageCrop(effect.image)
        }
    }
  ```
second points :
why we think about the navigation to another feature it something not releated to the navigation of Dukan feature , until the infra team decide that we can discuss about how can we integrate it in our feacture

haideraljawasim: the second point , we will handle navigation between features for example from the cart to wallet its not infra responsibility , i suggest to keep this refactor until we find a way to navigate to other features also

ahmedNaser7: we are still in the same point , if we can handle the navigation to another feature using effect , we can handle it using the approach of navigation in view model

haideraljawasim: nah that’s not entirely accurate. If it were we wouldn’t have handled the navigation here with this callback, because the same scenario may happens when navigating to another feature.

ahmedNaser7: still didnt get your point , what is your opinion about how we will navigate to another feacture ?

YoussefmSaber (approved):

AbdelrhmanAyman20470 (approved):

ElsayedMagdy122 (approved):

PR #166, opened 2025-09-25
feature/CHT-36-storing-messages-in-room

Mohamed-Elhanafy:
This pull request introduces Room database integration for chat message persistence in the `core_chat_data` module, supporting both Android and iOS targets. It adds the necessary dependencies, implements the database schema and DAO, and configures dependency injection to provide access to the database and its DAO throughout the application.


core_chat_data/src/androidMain/kotlin/net/thechance/mena/core_chat/data/di/dataProviderModule.android.kt
```diff
@@ -3,19 +3,29 @@ package net.thechance.mena.core_chat.data.di
...
     val context = this.get<Context>()
     return createDataStore {
         context.filesDir.resolve(dataStoreName).absolutePath
     }
+}
+
+actual fun Scope.getDatabaseBuilder(): RoomDatabase.Builder<ChatDatabase> {
+    val context = this.get<Context>()
+    val appContext = context.applicationContext
+    val dbFile = appContext.getDatabasePath("chat_database.db")
```
Joseph-Sameh-0: can you extract the name into constants in common main?

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/database/entity/MessageEntity.kt
```diff
@@ -0,0 +1,25 @@
...
+package net.thechance.mena.core_chat.data.database.entity
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+import kotlin.time.ExperimentalTime
+
+@Entity
+data class MessageEntity(
+    @PrimaryKey(autoGenerate = true) val id: Long,
```
Joseph-Sameh-0: the id will be of type UUID or what do you think ?

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/database/MessageStatusConverter.kt
```diff
@@ -0,0 +1,14 @@
...
+import androidx.room.TypeConverter
+import net.thechance.mena.core_chat.data.database.entity.MessageStatus
+
+class MessageStatusConverter {
+
+    @TypeConverter
+    fun fromMessageStatus(status: MessageStatus): String = status.name
+
+    @TypeConverter
+    fun toMessageStatus(status: String): MessageStatus = MessageStatus.valueOf(status)
```
Joseph-Sameh-0: this valueOf can throw 
you can use RunCatching with get or default

core_chat_data/src/iosMain/kotlin/net/thechance/mena/core_chat/data/di/dataProviderModule.ios.kt
```diff
@@ -31,3 +34,19 @@ actual fun Scope.createSettingsDataStore(): DataStore<Preferences>{
     }
 }
 
+actual fun Scope.getDatabaseBuilder(): RoomDatabase.Builder<ChatDatabase> {
+    val dbFilePath = documentDirectory() + "/chat_database.db"
```
Joseph-Sameh-0: same with the constant you will use in commonMain

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/database/MessageStatusConverter.kt
Mohamed-Elhanafy: I don't think we need the UUID functinality so what about to make it string and save the uuid of the message that coming from the backend ?

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/database/entity/MessageEntity.kt
Mohamed-Elhanafy: I don't think we need the UUID functinality so what about to make it string and save the uuid of the message that coming from the backend ?

Joseph-Sameh-0: is this save the coming from the backend or the pending until send ?

Joseph-Sameh-0 (approved):

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/database/dao/MessageDao.kt
```diff
@@ -0,0 +1,29 @@
...
+    fun getMessagesByChat(chatId: String): Flow<List<MessageEntity>>
+
+    @Query("UPDATE MessageEntity SET message_status = :status WHERE id = :id")
+    suspend fun updateMessageStatus(id: String, status: MessageStatus)
+
+    @Query("DELETE FROM MessageEntity WHERE id = :id")
+    suspend fun deleteMessage(id: String)
+
+    @Query("DELETE FROM MessageEntity WHERE id in (:ids)")
+    suspend fun deleteAllMessages(ids: List<String>)
```
AzizaHelmy: why we need this functionalty now?

AzizaHelmy (approved):

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/database/dao/MessageDao.kt
Mohamed-Elhanafy: Instead of deleting messages one by one, this method allows deleting multiple messages in a single database operation, which is more efficient I think we need it now becus there can be more than one message with faild state when we send it

omer1998 (approved):
great work both of you, but i wonder with this implementation does it support migration because the database relation might change in the future, and also do we need to consider this step now ?

AzizaHelmy (approved):

core_chat_data/build.gradle.kts
```diff
@@ -56,9 +68,15 @@ kover.reports {
...
-            "*.di",
-            "*.dto",
+        packages("*.di", "*.dto")
+       classes(
+            "*.*Database*",
+            "*.*DatabaseConstructor*",
+            "*.*Converter*",
+            "*.*Dao*",
+            "*.*Entity*",
+            "*.*MessageStatus*"
```
ZainabAlnajjar: you can exclude the package instead of the classes

ZainabAlnajjar (approved):

PR #168, opened 2025-09-25
feature/WLT-32-transaction-history-shared-structure

gehadepada:
add transaction entity


hagarabobakr (approved):

MennaSrour (approved):

Ahmed7osny1 (approved):

wallet_domain/src/commonMain/kotlin/net/thechance/mena/wallet/domain/entity/Transaction.kt
```diff
@@ -0,0 +1,24 @@
+package net.thechance.mena.wallet.domain.entity
+
+data class Transaction(
+    val id: String,
+    val createdAt: String,
```
nourelhodaahmed: I think make it date time is better than string

wallet_domain/src/commonMain/kotlin/net/thechance/mena/wallet/domain/entity/Transaction.kt
```diff
@@ -0,0 +1,24 @@
...
+)
+enum class TransactionType {
+    SENT,
+    RECEIVED,
+    ONLINE_PURCHASE
+}
+enum class TransactionStatus{
+    SUCCESS,
+    FAIL,
+    ALL
```
nourelhodaahmed: What `all` is for?

gehadepada: to get all transactions

Ahmed7osny1: We can create two enums 
One for status of transaction response
Two for transaction filter like TransactionStatusFilter

nourelhodaahmed: this should define the transaction status only that is either success or fail 
as all is only used in filtering  so i think it should be handled there not in here

Ahmed7osny1: You are right can be two enums as previous comment

gehadepada: okay  what is common between us just contain fail & success

Ahmed7osny1: Yes, just u want to create 
TransactionStatus take success and fail 
And 
TransactionFilterStatus take all, success and fail

abdallahFc (approved):

PR #178, opened 2025-09-26
Feature/TRN-32-update trend logic

fatimafaisalghazi:
Added updateReelById function to send a PUT request for updating reel data (description + category IDs).

Created UpdateReelRequestDTO to represent the request body and send it via setBody.

Used toEntity() to map the response from ReelDto to Reel.

Added parseDateStringOrNull util function to handle invalid date strings safely.


trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/util/mapperExtensions.kt
```diff
@@ -0,0 +1,15 @@
+package net.thechance.mena.trends.data.util
+
+import kotlinx.datetime.LocalDateTime
+
+fun Int?.orZero(): Int = this ?: 0
+
+
+fun String?.parseDateStringOrNull(): LocalDateTime? {
+    if (this.isNullOrBlank()) return null
+    return try {
```
ahmeddsamii: since we will not use the exception here what about using runCatching? it will be better 

replace try catch with this version

return runCatching {
LocalDateTime.parse(this)
}.getOrNull()

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/mapper/ReelMapper.kt
```diff
@@ -1,20 +1,20 @@
...
-        viewsCount = viewsCount ?: 0,
-        createdAt = LocalDateTime.parse(createdAt),
+        id = id.orEmpty(),
+        thumbnailUrl = reelImageUrl.orEmpty(),
+        videoUrl = videoUrl.orEmpty(),
+        description = description.orEmpty(),
+        likesCount = likesCount.orZero(),
+        viewsCount = viewsCount.orZero(),
+        createdAt = createdAt.parseDateStringOrNull(),
         categories = categories?.mapNotNull { it.toEntity() } ?: emptyList()
```
ahmeddsamii: instead of ?: emptyList() u can use .orEmpty()

trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/repository/ReelsRepository.kt
```diff
@@ -5,4 +5,5 @@ import net.thechance.mena.trends.domain.entity.Reel
 interface ReelsRepository {
     suspend fun deleteReelById(id: String)
     suspend fun getAllReels(pageNumber : Int): List<Reel>
+    suspend fun updateReelById(id: String,description: String,categoryIds: List<String>): Reel
```
ahmeddsamii: ctrl + alt + L 👀

ahmeddsamii (commented):
Great job, check the comments please

trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/repository/ReelsRepository.kt
```diff
@@ -4,5 +4,6 @@ import net.thechance.mena.trends.domain.entity.Reel
 
 interface ReelsRepository {
     suspend fun deleteReelById(id: String)
-    suspend fun getAllReels(pageNumber : Int): List<Reel>
+    suspend fun getAllReels(pageNumber: Int): List<Reel>
+    suspend fun updateReelById(id: String, description: String, categoryIds: List<String>): Reel
```
HendSayed25: why update reel function return reel ?

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/util/mapperExtensions.kt
```diff
@@ -0,0 +1,13 @@
+package net.thechance.mena.trends.data.util
+
+import kotlinx.datetime.LocalDateTime
+
+fun Int?.orZero(): Int = this ?: 0
```
HendSayed25, line 5: Can explain to me where will use this function ,why need it ?

trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/repository/ReelsRepository.kt
fatimafaisalghazi: because
1 - this is the response from the backend
2 - you just will update the description and select a category, and get a reel with these new values
so return reel

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/util/mapperExtensions.kt
fatimafaisalghazi: We need it when we want to make a default value of an integer val instead of using the Elvis operator,
it  same as `orEmpty`.

trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/repository/ReelsRepository.kt
HendSayed25: okay this function I think only responsible for update Trend to be published in main screen 
so I will get this new trend with all trends in another call 
but now if return here the updated reel , where I will use it ?

ahmeddsamii (approved):

trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/repository/ReelsRepository.kt
ahmeddsamii: it's okay to return the reel or not
@HendSayed25

HendSayed25: I Think in mobile we don't need it , we can replace this to ensure if update successfully or not by making the function return boolean and check response of back if the reel has the same data I send then return true else return false 
What's your opinion about this ? 
@ahmeddsamii

ahmeddsamii: if the update doesn't happen it will return an exception, we don't need boolean

HendSayed25 (approved):
Great work 👏👏

ahmedmahgoub0 (approved):
Good work👏

PR #180, opened 2025-09-26
test/FTH-36-bookmark-view-model

Mostafa-alsaygh:
In this PR  I add a test for the bookmark Viewmodel

### DIFF
--- a/faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
+++ b/faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
@@ -43,7 +43,7 @@ class BookmarkViewModelTest {
         verifySuspend { repository.getAllAyahBookmarks() }
     }
 
-
+    @OptIn(ExperimentalCoroutinesApi::class)
     @Test
     fun `onDeleteBookmarkClick should remove bookmark from state when repository succeeds`() =
         runTest {
@@ -53,6 +53,7 @@ class BookmarkViewModelTest {
             val viewModel = BookmarkViewModel(repository)
             viewModel.onDeleteBookmarkClick(1)
 
+            advanceUntilIdle()
             viewModel.uiState.test {
                 val state = awaitItem()
                 assertTrue(state.bookmarks.isEmpty())
--- a/faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
+++ b/faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
@@ -97,26 +97,25 @@ class BookmarkViewModelTest {
         }
     }
 
-}
-
-object FakeData {
-    @OptIn(ExperimentalTime::class)
-    val fakeBookmarks = listOf(
-        AyahBookmark(
-            id = 1,
-            surah = Surah(
+    private companion object FakeData {
+        @OptIn(ExperimentalTime::class)
+        val fakeBookmarks = listOf(
+            AyahBookmark(
                 id = 1,
-                order = Surah.SurahOrder.AlFatihah,
-                name = "Al-Fatihah",
-                ayahCount = 7,
-                isMakkia = true
-            ),
-            ayah = Ayah(
-                number = 1,
-                surahId = 1,
-                content = "بسم الله الرحمن الرحيم"
-            ),
-            createdAt = Instant.DISTANT_PAST
+                surah = Surah(
+                    id = 1,
+                    order = Surah.SurahOrder.AlFatihah,
+                    name = "Al-Fatihah",
+                    ayahCount = 7,
+                    isMakkia = true
+                ),
+                ayah = Ayah(
+                    number = 1,
+                    surahId = 1,
+                    content = "بسم الله الرحمن الرحيم"
+                ),
+                createdAt = Instant.DISTANT_PAST
+            )
         )
-    )
+    }
 }
--- a/faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
+++ b/faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
@@ -74,7 +74,7 @@ class BookmarkViewModelTest {
     }
 
     @Test
-    fun `onBackClick should should navigate back`() = runTest {
+    fun `onBackClick should navigate back`() = runTest {
         val viewModel = BookmarkViewModel(repository)
 
         viewModel.uiEffect.test {
--- a/faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
+++ b/faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
@@ -26,18 +26,15 @@ class BookmarkViewModelTest {
 
     @Test
     fun `init should load bookmarks when viewModel is created`() = runTest {
-
         everySuspend { repository.getAllAyahBookmarks() } returns FakeData.fakeBookmarks
 
-
         val viewModel = BookmarkViewModel(repository)
-
-
         viewModel.uiState.test {
             val state = awaitItem()
             assertEquals(1, state.bookmarks.size)
             assertEquals(1, state.bookmarks.first().bookmarkId)
         }
+
         verifySuspend { repository.getAllAyahBookmarks() }
     }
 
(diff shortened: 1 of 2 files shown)


yasserahmed10 (approved):

faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
```diff
@@ -0,0 +1,121 @@
...
+                assertTrue(state.bookmarks.isEmpty())
+            }
+            verifySuspend { repository.deleteAyahBookmark(1) }
+        }
+
+
+    @Test
+    fun `onDeleteBookmarkClick should keep bookmarks when repository fails`() = runTest {
+        everySuspend { repository.getAllAyahBookmarks() } returns fakeBookmarks
+        everySuspend { repository.deleteAyahBookmark(1) } throws RuntimeException("delete failed")
```
fatmahgazy: I think we don’t need to handle the error for now since there’s no implementation for the exception yet what do you think?

faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
```diff
@@ -0,0 +1,121 @@
...
+        viewModel.uiState.test {
+            val state = awaitItem()
+            assertEquals(1, state.bookmarks.size)
+            assertEquals(1, state.bookmarks.first().bookmarkId)
+        }
+
+        verifySuspend { repository.getAllAyahBookmarks() }
+    }
+
+    @OptIn(ExperimentalCoroutinesApi::class)
```
fatmahgazy: You can add it on `class BookmarkViewModelTest` instead add it to each function what do you think ?

fatmahgazy (commented):
Greate Work Mostafa 👏👏

faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
Mostafa-alsaygh: do u need to suggest to remove  this test function ?

faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
Mostafa-alsaygh: i will see it

faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
```diff
@@ -0,0 +1,121 @@
...
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlin.time.ExperimentalTime
+import kotlin.time.Instant
+
+class BookmarkViewModelTest {
+
+    private val repository: BookmarkRepository = mock(MockMode.autofill)
+
+    @OptIn(ExperimentalCoroutinesApi::class)
```
marwanqashwa: please move this annotation to the top of the class

faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
```diff
@@ -0,0 +1,121 @@
...
+    @Test
+    fun `init should load bookmarks when viewModel is created`() = runTest {
+        everySuspend { repository.getAllAyahBookmarks() } returns fakeBookmarks
+
+        val viewModel = BookmarkViewModel(repository)
+
+        advanceUntilIdle()
+        viewModel.uiState.test {
+            val state = awaitItem()
+            assertEquals(1, state.bookmarks.size)
```
marwanqashwa: this magic numbers doesn't  express anything. you could check against the fake bookmarks size, or create FakeBookmarksUi and compare with it directly

faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
```diff
@@ -0,0 +1,121 @@
...
+
+            val viewModel = BookmarkViewModel(repository)
+            viewModel.onDeleteBookmarkClick(1)
+
+            advanceUntilIdle()
+            viewModel.uiState.test {
+                val state = awaitItem()
+                assertTrue(state.bookmarks.isEmpty())
+            }
+            verifySuspend { repository.deleteAyahBookmark(1) }
```
marwanqashwa: this one can be const BOOKMARK_ID

faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
```diff
@@ -0,0 +1,121 @@
...
+        viewModel.onDeleteBookmarkClick(1)
+
+        viewModel.uiState.test {
+            val state = awaitItem()
+            assertEquals(1, state.bookmarks.size)
+        }
+    }
+
+    @Test
+    fun `onBackClick should emit NavigateBack effect`() = runTest {
```
marwanqashwa: suggestion you could make it more english like  onBackClick should navigate back

faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
```diff
@@ -0,0 +1,121 @@
...
+        val viewModel = BookmarkViewModel(repository)
+
+        advanceUntilIdle()
+        viewModel.uiState.test {
+            val state = awaitItem()
+            assertEquals(1, state.bookmarks.size)
+            assertEquals(1, state.bookmarks.first().bookmarkId)
+        }
+
+        verifySuspend { repository.getAllAyahBookmarks() }
```
FaresM0hamed: You don't need to use verify here , Once you assert some thing and it return true or false , that's mean you already hit this line , So remove this verify and it will be good if you make one assertion by function .

faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    @Test
+    fun `onDeleteBookmarkClick should remove bookmark from state when repository succeeds`() =
+        runTest {
+            everySuspend { repository.getAllAyahBookmarks() } returns fakeBookmarks
```
FaresM0hamed, line 47: Here what is the purpose for this stub here , Do we need it ?

Mostafa-alsaygh: to check if there is a data after the deletion in this line `assertTrue(state.bookmarks.isEmpty())`

faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
Mostafa-alsaygh: it is ok when we work to the viewmodel we will make it like the test

Mostafa-alsaygh: @FaresM0hamed what is your suggestion

faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
FaresM0hamed: I think we could test it with verifySuspend only and don't need to assert here

faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
```diff
@@ -0,0 +1,122 @@
...
+        viewModel.onDeleteBookmarkClick(BOOKMARK_ID)
+
+        viewModel.uiState.test {
+            val state = awaitItem()
+            assertEquals(1, state.bookmarks.size)
+        }
+    }
+
+    @Test
+    fun `onBackClick should should navigate back`() = runTest {
```
FaresM0hamed: Typo here by repeating `should`

faith_presentation/src/commonTest/kotlin/presentation/feature/bookmark/BookmarkViewModelTest.kt
Mostafa-alsaygh: but the fun do two things call the repo then update the state and we need to check both

Mostafa-alsaygh: as the name of function `'should remove bookmark from state when repository succeeds'`

Al-Taie (approved):

FaresM0hamed (approved):

PR #194, opened 2025-09-27
Feature/IDN-73 replace image magick with static icons

eslam-99:
### Summary
In this PR, I removed image magick completely and added static icons for each environment for Android & IOS

### Changes
- Removed build script of Android and replced with static icons
- Removed build script of IOS and replced with static icons
- Removed image magick install script from CI


iBareq (changes requested):
please use MENA icon instead of that goof android bot

iBareq (approved):

PR #198, opened 2025-09-27
add main screen navigation

AimanYosofiUlt:
add main screen navigation


iBareq (approved):
Hello fellas, I will approve this PR and merge it without check
later I may refactor it to resolve the blocker that we are causing for other teams

PR #208, opened 2025-09-27
WLT-35-add-modules-to-di

Al-Qassim:
add wallet modules to app di


MennaSrour (approved):

iBareq:
closed due to having all modules di in the app after some refactoring

PR #217, opened 2025-09-27
fix/FTH-36-time-ago-recomposition-issue

Mostafa-alsaygh:
in this PR i fix the implement of bookmark creation extension time as 'time ago' from the mapper instead of the screen to solve recomposition issue


faith_presentation/src/commonMain/kotlin/net/thechance/mena/faith/presentation/extensions/timeFormatingHelper/getTimeAgo.kt
```diff
@@ -64,4 +69,4 @@ private const val ONE_HOUR_IN_MINUTES = 60
 private const val ONE_DAY_IN_HOURS = 24
 private const val ONE_WEEK_IN_DAYS = 7
 private const val ONE_MONTH_IN_DAYS = 30
-private const val ONE_YEAR_IN_DAYS = 365
\ No newline at end of file
+private const val ONE_YEAR_IN_DAYS = 365
```
fatmahgazy, line 72: This is just a suggestion what do you think if we use a `companion object` here?

fatmahgazy (approved):
Greate  work Mostafa👏👏

faith_presentation/src/commonMain/kotlin/net/thechance/mena/faith/presentation/extensions/timeFormatingHelper/getTimeAgo.kt
Mostafa-alsaygh: we cant cause it is not a class so no other way

fatmahgazy: Aha, you mean that these values are general and used in a general way, not belonging to a specific class. Yes, you’re right indeed.

Mostafa-alsaygh: no but this is a fun not a class to create a companion object

fatmahgazy: Yes, I got you and with that, you can also use them in a general way.
 I just mistakenly saw them inside a class instead of at the top level of the file thanks for clarification

marwanqashwa (approved):

faith_presentation/src/commonMain/kotlin/net/thechance/mena/faith/presentation/extensions/timeFormatingHelper/TimeAgo.kt
```diff
@@ -0,0 +1,8 @@
+package net.thechance.mena.faith.presentation.extensions.timeFormatingHelper
+
+data class TimeAgo(
+    val amount: Int = noValue,
+    val unit: TimeUnit = TimeUnit.SECONDS
+)
+
+const val noValue = 0
```
Al-Taie: Remove this, it is redundant.

faith_presentation/src/commonMain/kotlin/net/thechance/mena/faith/presentation/extensions/timeFormatingHelper/TimeUnit.kt
```diff
@@ -0,0 +1,5 @@
+package net.thechance.mena.faith.presentation.extensions.timeFormatingHelper
+
+enum class TimeUnit {
+    SECONDS, MINUTES, HOURS, DAYS, WEEKS, MONTHS, YEARS
```
Al-Taie: It would be easier to read if it's separated by lines.
```suggestion
    SECONDS, 
    MINUTES, 
    HOURS, 
    DAYS, 
    WEEKS, 
    MONTHS, 
    YEARS
```

Al-Taie (approved):

PR #218, opened 2025-09-27
WLT-34-make-navigation

Al-Qassim:
add navigation for the wallet feature

https://github.com/user-attachments/assets/e06344be-3852-4aeb-a468-9e0a807d3c10


wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/navigation/Routes.kt
MuhammedEdrees, line 1: I think it would be better to make `WalletRoute` a sealed class to be more controlled.

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/navigation/Routes.kt
```diff
@@ -0,0 +1,30 @@
...
+data object TransactionsHistoryScreenRoute: WalletRoute()
+
+@OptIn(ExperimentalUuidApi::class)
+@Serializable
+data class TransactionDetailsScreenRoute(
+    val id: String
+): WalletRoute() {
+    init {
+        Uuid.parse(id)
+    }
```
MuhammedEdrees, line 23: Is this for validation?

MuhammedEdrees (approved):

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/navigation/Routes.kt
Al-Qassim: yes

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/export_transactions/ExportTransactionsScreen.kt
```diff
@@ -0,0 +1,62 @@
...
+            .fillMaxSize()
+            .padding(16.dp),
+        verticalArrangement = Arrangement.spacedBy(
+            space = 16.dp,
+            alignment = Alignment.CenterVertically
+        ),
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+        Text(
+            text = "Export Transactions Screen",
```
Farah315, line 37: i think we can use here string resource

Al-Qassim: this is just a place holder in a fake screen, later this will be replaced with real screen

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/export_transactions/ExportTransactionsScreen.kt
```diff
@@ -0,0 +1,62 @@
...
+import net.thechance.mena.designsystem.presentation.theme.theme.Theme
+import kotlin.uuid.ExperimentalUuidApi
+
+@OptIn(ExperimentalUuidApi::class)
+@Composable
+fun ExportTransactionsScreen(
+    onNavigateBackClicked: () -> Unit,
+    navigateToVewTransactionStatement: () -> Unit
+) {
+    Column(
```
Farah315, line 25: can i ask why you didn't use Wallet Scaffold ?

Al-Qassim: this is just a fake screen, later this will be replaced with real screen

Ahmed7osny1 (approved):

Farah315 (approved):
great job qassim, there just a little thing like using string resource to we can easily translation our wallet and i think when we use scaffold would be faster to use topBar or snackBar,, etc

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/navigation/Routes.kt
MuhammedEdrees: I'm not sure if this is the best way to handle invalid uuid since it will crash the app

MuhammedEdrees: May be we can validate it inside the `composable<TransactionDetailsScreenRoute>` block and cancel the navigation if it's invalid?

PR #224, opened 2025-09-28
fix: force specific coverage percentage in codecov for any changed code.

ThoriaHamdy99:
# Changes:
- Add codecov.yml to put specific configuration to be used by codecov
- Add coverage 80% for all new changes on the PR.
- Let the codecov don't fail when the whole project coverage is less than expected.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,8 @@
+coverage:
+  status:
+    project:
+      default:
+        enabled: no
+    patch:
+      default:
+        target: 80%
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,7 +2,7 @@ coverage:
   status:
     project:
       default:
-        enabled: no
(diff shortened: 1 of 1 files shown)


eslam-99 (commented):
Great to split each module coverage percentage, but this will enforce same 80% to all testable modules
The percentage may vary on every module

ThoriaHamdy99:
> Great to split each module coverage percentage, but this will enforce same 80% to all testable modules The percentage may vary on every module

what I did:
module1: 80% 
modules2: 80%
As made at the first but if we need module1: 80% and mosule2: 60% for example so there is a solution I can make it but let's have a discussion on our meeting first.

AhmadKharfan (approved):

AimanYosofiUlt (approved):

shahad-97 (approved):

PR #227, opened 2025-09-28
Feature/trn 40 view model test

skaik-mo:
Inject CoroutineDispatcher into CategoryPickViewModel

Add unit tests for CategoryPickViewModel

loadCategories starts loading initially.
loadCategories successfully returns categories when the repository provides data.
loadCategories throws an exception when the repository call fails.
loadCategories ends loading after completion.
onCategoryClick toggles the selection state of a category.
onBackClick sends a NavigateBack effect.
onNextClick successfully calls updateUserInterestedCategories and sends a NavigateToTrends effect.
onNextClick throws an exception if the repository call fails.


HendSayed25 (approved):

iosApp/Podfile.lock
```diff
@@ -0,0 +1,16 @@
+PODS:
+  - MapLibre (6.17.1)
+
+DEPENDENCIES:
+  - MapLibre (= 6.17.1)
+
+SPEC REPOS:
```
ahmeddsamii: I think we shouldn't change these files, yeah?

skaik-mo: This file should be added to `.gitignore`

ahmeddsamii: Okay then revert any changes not related to the pr

ahmedmahgoub0 (approved):

ahmeddsamii (approved):

mohamedd-hassan (approved):

PR #255, opened 2025-10-01
Feature/TRN-41 upload video screen ui

iZyadNasser:
## This PR introduces the video upload screen for the trends upload new video user story

- You can open the system video picker and choose a video to upload
- There is a validation for the video, it must be 100MB or less in size, and 1 minute or less in duration
- You can follow your upload status and get instant feedback if the upload failed at anytime

 _Note: the upload video will always give you failed status right now because the backend endpoint has not been merged nor deployed yet_

photo_2025-10-01_20-35-41
photo_2025-10-01_20-36-37
photo_2025-10-01_20-35-44


trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/entity/UploadReelProgress.kt
```diff
@@ -1,7 +1,7 @@
 package net.thechance.mena.trends.domain.entity
 
 data class UploadReelProgress(
-    val reelId: String,
+    val reelId: String = "",
```
ahmeddsamii, line 4: why is that?

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/upload_reel/UploadReelScreenState.kt
```diff
@@ -0,0 +1,26 @@
...
+
+data class UploadReelScreenState(
+    val selectedFile: FileUiState = FileUiState(),
+    val thumbnail: PlatformFile? = null,
+    val uploadingTrendState: UploadingTrendState = UploadingTrendState.IDLE,
+    val uploadedBytes: Long = 0,
+    val isNextButtonEnabled: Boolean = false,
+    val isNextButtonLoading: Boolean = false,
+    val errorState: ErrorState? = null,
+    val trendId: String? = null
```
mohamedd-hassan, line 15: in my opinion we should remove the trendId from here since its not a state that appears to the user, I would suggest sending it with the effect

mohamedd-hassan (approved):

trends_domain/src/commonMain/kotlin/net/thechance/mena/trends/domain/entity/UploadReelProgress.kt
iZyadNasser: we only need it in the last emit from the flow, I gave it a default value, so we don't have to pass it with every emit, only the last

ahmeddsamii (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/UploadVideoCard.kt
```diff
@@ -37,7 +37,7 @@ import org.jetbrains.compose.ui.tooling.preview.Preview
 @Composable
 internal fun UploadVideoCard(
     modifier: Modifier = Modifier,
-    thumbnail: Painter? = null,
+    thumbnail: PlatformFile? = null,
```
ahmedmahgoub0, line 40: I think we don't need thumbnail to be of type `PlatformFile`, after extracting it it will be just a byte array or bitmap

ahmedmahgoub0 (approved):

PR #256, opened 2025-10-01
Test/ FTH 53 add hash check for Surah data class test

marwanqashwa:
Added a new test class `SurahTest` to verify the integrity of the `Surah.kt` file.
This test calculates the MD5 hash of `Surah.kt` and compares it against a known hash value.
This ensures that the `Surah` data class, which represents critical domain data, is not accidentally modified.

### DIFF
--- a/faith_domain/src/jvmTest/kotlin/net/thechance/mena/faith/domain/entity/SurahTest.kt
+++ b/faith_domain/src/jvmTest/kotlin/net/thechance/mena/faith/domain/entity/SurahTest.kt
@@ -14,12 +14,12 @@ class SurahTest {
         assertEquals(
             FILE_HASH,
             actualHash,
-            message = "Surah.kt file has been modified!${actualHash}\n Please check the changes."
+            message = "Surah.kt file has been modified! new hash:${actualHash}\n Please check the changes."
         )
     }
 
     private companion object {
-        const val FILE_HASH = "9d8cb4694b0e7e74899c8d62a5cedc97"
+        const val FILE_HASH = "3ee45be83f266197549c3a15534bfe18"
         const val SURAH_FILE_PATH =
             "src/commonMain/kotlin/net/thechance/mena/faith/domain/entity/Surah.kt"
     }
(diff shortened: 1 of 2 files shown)


BassantM (commented):
Good Job @marwanqashwa, just a suggestion 
It would be good to add a check to verify the file exists before calculating its MD5 hash.
This will help avoid unexpected `FileNotFoundException` errors if the file is moved or deleted. What do u think

faith_domain/src/jvmTest/kotlin/net/thechance/mena/faith/domain/entity/SurahTest.kt
```diff
@@ -0,0 +1,42 @@
...
+
+class SurahTest {
+
+    @Test
+    fun `verify that Surah data class has not been modified`() {
+        val actualHash = calculateFileMd5Hash(File(SURAH_FILE_PATH))
+        assertEquals(
+            FILE_HASH,
+            actualHash,
+            message = "Surah.kt file has been modified! actual hash:${actualHash}\n Please check the changes."
```
fatmahgazy, line 17: `message = "Surah.kt file has been modified! actual hash:${actualHash}\n Please check the changes."`
I think you can add it to companion object to be test clear and readable what do you think?

Overall good work Marwan 👏👏

```

marwanqashwa:
> Good Job @marwanqashwa, just a suggestion 
> It would be good to add a check to verify the file exists before calculating its MD5 hash.
> This will help avoid unexpected `FileNotFoundException` errors if the file is moved or deleted. What do u think

I think it's okay to have a FileNotFoundException since this is not a production environment. If the exception is thrown, the test will just fail. What do you think?

faith_domain/src/jvmTest/kotlin/net/thechance/mena/faith/domain/entity/SurahTest.kt
marwanqashwa: I think it's ok because we concatenate the actual hash in the message what do you think?

BassantM: > ```
>         `message = "Surah.kt file has been modified! actual hash:${actualHash}\n Please check the changes."`
> ```
> 
> I think you can add it to companion object to be test clear and readable what do you think?
> 
> Overall good work Marwan 👏👏

I think it is clear, as it is only one test case, what is ur opinion @fatmahgazy

BassantM (approved):

fatmahgazy (approved):

yusufnasserdev (approved):

PR #280, opened 2025-10-03
Feature/TRN-52-Implement reel deletion in UploadReelViewModel

ahmeddsamii:
Refactored `onDeleteVideoClick` to call the `deleteReelById` repository method and handle the success and error states accordingly.

This change introduces a new private `deleteVideo` function and adds corresponding unit tests to verify the new deletion logic, including success and failure scenarios.

image

### DIFF
--- a/trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/screen/upload_reel/UploadTrendViewModelTest.kt
+++ b/trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/screen/upload_reel/UploadTrendViewModelTest.kt
@@ -2,7 +2,6 @@ package net.thechance.mena.trends.presentation.screen.upload_reel
 
 import app.cash.turbine.test
 import assertk.assertThat
-import assertk.assertions.isEmpty
 import assertk.assertions.isEqualTo
 import assertk.assertions.isFalse
 import assertk.assertions.isNull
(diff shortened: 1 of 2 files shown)


mohamedd-hassan (approved):

fatimafaisalghazi (approved):

trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/screen/upload_reel/UploadTrendViewModelTest.kt
```diff
@@ -61,60 +58,67 @@ class UploadTrendViewModelTest: TestExtensions() {
...
 
     @Test
-    fun `onRetrieveVideo should update error state with FileTooLarge if size is not valid`() = runTest(testDispatcher) {
-        viewModel.onRetrieveVideo(fileWithInvalidSize, ::defaultReadBytes)
-        advanceUntilIdle()
-
-        viewModel.state.test {
-            assertThat(awaitItem().errorState).isEqualTo(ErrorState.FileTooLarge)
-            cancelAndIgnoreRemainingEvents()
+    fun `onRetrieveVideo should update error state with FileTooLarge if size is not valid`() =
```
HendSayed25: I think the keyword( is not valid) contains different meanings , but in this case you test only when size of file exceed the max limit we determined so I suggest rename the test to be : 
```suggestion
    fun `onRetrieveVideo should update error state with FileTooLarge when file's size exceed the max limit`() =
```

HendSayed25 (approved):

trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/screen/upload_reel/UploadTrendViewModelTest.kt
ahmeddsamii: i didn't write this test case i just reformat the entire file, but okay

iZyadNasser (approved):

PR #287, opened 2025-10-03
Fix Overlays Visibility State

nabilBouzineDev:
This PR passes the `isVisible` boolean to the `content` composable lambda for overlays like `bottomSheet` and `dialog` within the `ScaffoldScope`. This allows the overlay content to be aware of its own visibility state.

The affected components:

- `TransactionFilterBottomSheet` (wallet_presentation)
- `CountryPicker` (identity_presentation module)


Al-Qassim (approved):

design_system/src/commonMain/kotlin/net/thechance/mena/designsystem/presentation/component/bottomSheet/BottomSheet.kt
```diff
@@ -72,6 +72,7 @@ enum class BottomSheetValue {
 fun ScaffoldScope.BottomSheet(
```
poula22, line 72: apply deprecation approach with message you should pass is visible to new function as its now mandatory field not optional

design_system/src/commonMain/kotlin/net/thechance/mena/designsystem/presentation/component/bottomSheet/BottomSheet.kt
```diff
@@ -259,8 +264,9 @@ private fun SimpleBottomSheetPreview() {
     MenaTheme {
         Scaffold(
             overlays = {
-                bottomSheet(isVisible = visible) {
+                bottomSheet(isVisible = visible) { isVisible ->
                     BottomSheet(
+
```
poula22: you should add isVisible = isVisible

design_system/src/commonMain/kotlin/net/thechance/mena/designsystem/presentation/component/bottomSheet/BottomSheet.kt
nabilBouzineDev: I don't think we need it in this case since I've updated all the components that use this component

poula22 (approved):

AimanYosofiUlt (approved):

ABDULLAHHG (approved):

AhmadKharfan (approved):

shahad-97 (approved):

iBareq (approved):

PR #307, opened 2025-10-05
Fix/wlt 59 increase test coverage

Farah315:
hi there, is PR are fix broken test in view model and increase transaction history view model test , and improve readability in it  
☝🏻
image


nourelhodaahmed (approved):

MennaSrour (approved):

gehadepada (approved):

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/transaction_history/TransactionHistoryViewModel.kt
```diff
@@ -32,7 +33,8 @@ import kotlin.uuid.Uuid
 @OptIn(ExperimentalUuidApi::class)
 @KoinViewModel
 class TransactionHistoryViewModel(
-    @Provided private val transactionRepository: TransactionRepository
+    @Provided private val transactionRepository: TransactionRepository,
+    private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO
```
Joseph-Sameh-0, line 37:
```suggestion
    private val dispatcher: CoroutineDispatcher = Dispatchers.IO
```

I think it's better to name it just dispatcher because we change it in the test to testDispatcher so it's not always ioDispatcher

Joseph-Sameh-0 (approved):

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/transaction_history/TransactionHistoryViewModel.kt
Farah315: @Joseph-Sameh-0  yah its sound good but here i rename as ioDispatcher cuz in any ViewModel here we pass it as io dispatcher, and in test as usual we rename it test dispatcher

Ahmed7osny1 (approved):

PR #316, opened 2025-10-08
Feature/TRN - 60 Cover Missing Test Coverage

MahmoodTarek:
### What I did in this PR
- Added missing test cases and related test files to improve coverage.
- Refactored existing test cases for better readability and consistency.


trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/screen/category_publish/CategoryPublishViewModelTest.kt
```diff
@@ -54,53 +54,99 @@ class CategoryPublishViewModelTest {
...
-
-        verifySuspend {
-            reelsRepository.updateReelById(
-                id = "trendId",
-                description = "description",
-                categoryIds = listOf(sampleCategory.id)
-            )
+    fun `onPublishClick should call updateReelById with correct params when category is selected`() =
+        runTest {
+            every { categoryPublishArgs.trendId } returns TREND_ID
```
ahmeddsamii: you can move all every in BeforeTest block

trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/screen/category_publish/CategoryPublishViewModelTest.kt
```diff
@@ -54,53 +54,99 @@ class CategoryPublishViewModelTest {
...
+            every { categoryPublishArgs.description } returns DESCRIPTION
```
ahmeddsamii: same here and for all every in this file

trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerViewModelTest.kt
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    @Test
+    fun `navigateToCategories should navigate to categories screen when user categories are already set`() =
+        runTest {
+            viewModel.handleGetIsUserCategorySet(isUserCategorySet = false)
+            viewModel.effect.test {
+                viewModel.navigateToCategories()
+                assertThat(awaitItem()).isEqualTo(MainContainerEffect.NavigateToCategoryPick)
```
ahmeddsamii, line 65: the test name said that user already has categories, so here why you navigate to CategoryPick? i think you should navigate to TrendScreen direct

trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerViewModelTest.kt
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+
+    @Test
+    fun `navigateToCategories should navigate to trends screen when user categories are not set`() =
+        runTest {
+            viewModel.handleGetIsUserCategorySet(isUserCategorySet = true)
+            viewModel.effect.test {
+                viewModel.navigateToCategories()
+                assertThat(awaitItem()).isEqualTo(MainContainerEffect.NavigateToTrends)
```
ahmeddsamii, line 75: this should be instead of the previous test, i mean here should navigate to categoryPick, and the previous one should navigate to Trends

trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerViewModelTest.kt
```diff
@@ -0,0 +1,98 @@
...
+
+    @Test
+    fun `handleGetIsUserCategorySet should update state with isUserCategorySet`() = runTest {
+        viewModel.handleGetIsUserCategorySet(isUserCategorySet = true)
+        assertThat(viewModel.state.value.isCategoriesAlreadySelectedByUser).isEqualTo(true)
+    }
+
+    @Test
+    fun `loadCategories should update error state when repository throws exception`() = runTest {
+        everySuspend { repository.isCategoriesAlreadySelectedByUser() } throws Exception()
```
ahmeddsamii, line 87: you can move this to BeforeTest block

trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerViewModelTest.kt
```diff
@@ -0,0 +1,98 @@
...
+
+        val viewModel = MainContainerViewModel(repository, testDispatcher)
```
ahmeddsamii, line 89: and i think this one too can be in before test block

trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/screen/manage_my_trends/ManageTrendsViewModelTest.kt
```diff
@@ -51,8 +50,7 @@ class ManageTrendsViewModelTest {
     @Test
     fun `initialize view model should handle error state when getAllReels fails`() =
         runTest(testDispatcher) {
-            val errorMessage = "error"
-            everySuspend { repository.getAllReels(1) } throws Exception(errorMessage)
+            everySuspend { repository.getAllReels(1) } throws Exception()
```
ahmeddsamii: same here, you can move this to BeforeTest, and check the whole file

ahmeddsamii (changes requested):
great effort 👏, check the comments please and there is ReelRepositoryImpl, you missed some areas to cover

trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/utils/MockData.kt
```diff
@@ -1,14 +1,11 @@
...
-        "\n" +
-        "In the evenings, after the sun had set, she would return home carrying her precious books, counting the days until she could again sit among the pages and get lost in thrilling tales. As she explored these literary worlds, she discovered words that were unforgettable and letters that conveyed deep emotions. Every chapter she read or wrote opened new doors for her to think, explore, and create. Gradually, she built a small library in her room, filled with books, journals, and magazines, constantly adding new stories every day. She loved sharing what she read with her friends and family, discussing adventures of characters, plot twists, and lessons learned. \n" +
```
HendSayed25, line 6: why remove this ?

trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerViewModelTest.kt
MahmoodTarek: It’s just an `error simulation` for this specific test case, so I don’t need to set it up in the Before block for every test.

trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerViewModelTest.kt
MahmoodTarek: if I delete `viewmodel` initial from here and Just do it in `init` block  
The `state.test` block starts collecting before the coroutine in the `init` block finishes executing, so the error state doesn’t get updated in time, which causes the test to fail.

trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/screen/manage_my_trends/ManageTrendsViewModelTest.kt
MahmoodTarek: It’s just an error simulation for this specific test case, so I don’t need to set it up in the block for every test.

trends_presentation/src/commonTest/kotlin/net/thechance/mena/trends/presentation/utils/MockData.kt
MahmoodTarek: Because in mock data, we should only keep the data that’s shared across multiple files.
If the data is used only within the same file across multiple test cases, we keep it inside a companion object.
And if it’s used only in a single test case, we define it directly in the “given” section of that test.

MahmoodTarek: `largeNewDescription` is only used in a single test case, so I moved it to the “given” section of that specific test.
I also used Kotlin’s functional style:
`val largeNewDescription = "stanbul ".repeat(500)`
to generate the value dynamically instead of defining it as a static mock constant.

ahmeddsamii (approved):

norhanadel367 (approved):

ahmedmahgoub0 (approved):

PR #333, opened 2025-10-09
Feature/TRN 68 refactor components code

fatimafaisalghazi:
Add previews for all components.
Make the components internal in the trend module.
Remove duplicates by extracting them into a private component.

### DIFF
--- a/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/NextButton.kt
+++ b/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/NextButton.kt
@@ -1,19 +1,14 @@
 package net.thechance.mena.trends.presentation.shared.component
 
-import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
 import mena.trends_presentation.generated.resources.Res
 import mena.trends_presentation.generated.resources.next
-import net.thechance.mena.designsystem.presentation.component.button.Button
-import net.thechance.mena.designsystem.presentation.component.text.Text
+import net.thechance.mena.designsystem.presentation.component.button.PrimaryButton
 import net.thechance.mena.designsystem.presentation.theme.theme.MenaTheme
-import net.thechance.mena.designsystem.presentation.theme.theme.Theme
 import org.jetbrains.compose.resources.stringResource
 import org.jetbrains.compose.ui.tooling.preview.Preview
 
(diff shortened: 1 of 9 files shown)


trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/NextButton.kt
```diff
@@ -12,15 +12,17 @@ import mena.trends_presentation.generated.resources.Res
...
 
 @Composable
-fun NextButton(
-    onNextClick: () -> Unit,
+internal fun NextButton(
     isButtonEnabled: Boolean,
     isButtonLoading: Boolean,
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
+    onNextClick: () -> Unit
```
HendSayed25, line 20: Prefer the modifier is being in the last parameter

norhanadel367 (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/VideoLoadingCardItem.kt
```diff
@@ -148,60 +119,97 @@ private fun VideoInfoSection(
...
                 }
 
                 UploadReelScreenState.UploadingTrendState.IDLE -> {}
             }
         }
     }
+}
+
+@Composable
+private fun Cancel(onAction: (VideoAction) -> Unit) {
```
HendSayed25: I saw the cancel , delete and reload are the same but different only in modifier , I suggest Replace them by one function and pass the modifier and action to it

norhanadel367 (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/NextButton.kt
ahmeddsamii: it should be first optional parameter and here in this case it's okay if we kept the call back the last item

ahmeddsamii (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/NextButton.kt
HendSayed25: This is the only parameter take default value , I think it's better to put it in the first or in the end

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/VideoLoadingCardItem.kt
fatimafaisalghazi: check

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/NextButton.kt
```diff
@@ -12,15 +12,17 @@ import mena.trends_presentation.generated.resources.Res
...
 ) {
     Button(
```
ahmedmahgoub0: try to use the `PrimaryButton` from design system 
they had already designed the button for us, so we don't need to create it from scratch 
it will look like this:
```
@Composable
fun NextButton(
    onNextClick: () -> Unit,
    isButtonEnabled: Boolean,
    isButtonLoading: Boolean,
    modifier: Modifier = Modifier
) {
    PrimaryButton(
        modifier = modifier.fillMaxWidth(),
        text = stringResource(Res.string.next),
        onClick = onNextClick,
        isEnabled = isButtonEnabled,
        isLoading = isButtonLoading,
        contentPadding = PaddingValues(vertical = 13.dp)
    )
}
```

ahmedmahgoub0 (approved):
just note for upload components, I have changed them entirely in my pr 
Good Work 👏👏

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/NextButton.kt
fatimafaisalghazi: i did this becase some time we nedd use call back and pass a lampda in it 
like this 

` NextButton(
            isButtonEnabled = true,
            isButtonLoading = false
        ){
            //AAction
        }`
        what do u think ?
@HendSayed25

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/NextButton.kt
fatimafaisalghazi: thank u ahmed
image

image

ahmeddsamii (approved):

PR #338, opened 2025-10-09
replace FileSaver with specific PDF download

Ahmed7osny1:
https://thechance-workspace.slack.com/archives/C09AX4J62LR/p1760049025171889


wallet_presentation/src/iosMain/kotlin/net/thechance/mena/wallet/presentation/utils/pdfUtils.ios.kt
```diff
@@ -137,9 +143,64 @@ actual class PdfHandler {
...
+        return "${baseName}_$timestamp"
+    }
+
     private companion object {
         // Chosen as a good balance between rendering time and image sharpness
         const val IMAGE_SCALE = 1.67f
-
+        private const val DOWNLOAD_DIR_BASE = "Documents"
+        private const val DOWNLOAD_DIR_ROOT = "MENA"
     }
```
gehadepada: the companion object itself is private so I think no need to make values private

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/utils/pdfUtils.kt
```diff
@@ -6,4 +6,11 @@ import org.koin.core.annotation.Single
 expect class PdfHandler() {
     suspend fun splitToPagesOfPngs(pdfData: ByteArray): List<ByteArray>
     suspend fun sharePdf(pdfData: ByteArray, fileName: String)
+    suspend fun downloadPdf(pdfData: ByteArray, fileName: String): FileSaveResult
+}
+
+
+sealed class FileSaveResult {
```
gehadepada: delete extra line

wallet_presentation/src/commonTest/kotlin/net/thechance/mena/wallet/presentation/screen/export/ExportTransactionsViewModelTest.kt
```diff
@@ -648,12 +620,33 @@ class ExportTransactionsViewModelTest {
         }
     }
 
+    @Test
+    fun `downloadPdf returns success with file path`() = runTest {
```
gehadepada, line 624: I think it's better to give name to this text function like the remaining function for consistency

wallet_presentation/src/iosMain/kotlin/net/thechance/mena/wallet/presentation/utils/pdfUtils.ios.kt
```diff
@@ -137,9 +143,64 @@ actual class PdfHandler {
...
+        val timestamp = NSDate().timeIntervalSince1970.toLong() * 1000
+        return "${baseName}_$timestamp"
+    }
+
     private companion object {
         // Chosen as a good balance between rendering time and image sharpness
         const val IMAGE_SCALE = 1.67f
-
+        private const val DOWNLOAD_DIR_BASE = "Documents"
+        private const val DOWNLOAD_DIR_ROOT = "MENA"
```
Malak187: you can give them better naming as:
private const val DOCUMENTS_DIR_BASE = "Documents"
private const val APP_DOWNLOADS_FOLDER = "MENA"

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/component/SnackBarContainer.kt
```diff
@@ -51,7 +51,14 @@ fun SnackBarContainer(
...
-            message = snackBarState.messageRes?.let { stringResource(it) } ?: "",
+            message = when {
+                snackBarState.messageText != null && snackBarState.messageRes != null -> {
+                    "${stringResource(snackBarState.messageRes)} ${snackBarState.messageText}"
+                }
+                snackBarState.messageText != null -> snackBarState.messageText
+                snackBarState.messageRes != null -> stringResource(snackBarState.messageRes)
+                else -> ""
+            },
             leadingIcon = painterResource(leadingIcon),
```
Malak187: I don’t think the current approach for handling string resources in the Snackbar is actually the best. In CMP, you can safely use getString() without needing a context, so it’s cleaner to resolve the strings inside the ViewModel and store the message and title as String data type in the state instead of StringResource.

This might require some changes across our ViewModels, but it will make the Snackbar handling much simpler and more consistent.

This is the data class after changing the types from string resources to string:
```
data class SnackBarState(
    val isVisible: Boolean = false,
    val title: String? = null,
    val message: String? = null,
    val isSuccess: Boolean = true
)
 ```

And this is an example of the usage inside the view model in your case
```
is FileSaveResult.Success -> {
            val title = getString(Res.string.download_complete)
            val message = getString(Res.string.download_success) + isFileSaved.filePath
            showSnackBar(
                title = title,
                message = message,
                isSuccess = true
            )
        }
```
So what do you think?

**Review by Malak187 — CHANGES_REQUESTED** [review]

File: wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/component/SnackBarContainer.kt (hunk 5 above)
Comment (Ahmed7osny1) [reply wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/component/SnackBarContainer.kt]: Your approach will effect in all snack bar used, because all use message as string resource

File: wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/export/ExportTransactionsViewModel.kt [hunk 6]
```diff
@@ -395,7 +396,8 @@ class ExportTransactionsViewModel(
 
     private suspend fun showSnackBar(
         titleRes: StringResource,
-        messageRes: StringResource,
+        messageRes: StringResource? = null,
+        messageText: String? = null,
```
hagarabobakr: What is the use of this parameter?

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/component/SnackBarContainer.kt
MuhammedEdrees: I think it's a small refactor compared to having unneeded parameter in the snackbar state which can lead to misuse and will be much costier later on.

wallet_presentation/src/androidMain/kotlin/net/thechance/mena/wallet/presentation/utils/pdfUtils.android.kt
MuhammedEdrees, line 1: I think instead of returning `FileSaveResult` here we can return string in case of success or throw an exception in the case of error and handle it in the viewmodel for simplicity.

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/transaction_details/TransactionDetailsScreen.kt
```diff
@@ -170,13 +170,17 @@ private suspend fun onTransactionDetailsEffect(
...
                 onCaptureError()
             }
         }
 
         TransactionDetailsEffect.showErrorSnackBar -> {
-            showSnackBar(Res.string.error,Res.string.share_transaction_details_error_msg,false)
+            showSnackBar(
+                getString(Res.string.error),
+                getString(Res.string.share_transaction_details_error_msg),
+                false
```
MuhammedEdrees: For clarity, we can use named parameters here.

MuhammedEdrees (approved):
Well done 👏

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/export/ExportTransactionsViewModel.kt
```diff
@@ -312,48 +313,36 @@ class ExportTransactionsViewModel(
...
-                error = ErrorState.Unknown,
-                titleRes = Res.string.download_failed,
-                messageRes = Res.string.something_went_wrong,
+            showSnackBar(
+                title = getString(Res.string.download_failed),
+                message = getString(Res.string.something_went_wrong),
                 isSuccess = false
             )
         }
     }
```
MuhammedEdrees: I think this function could be smaller if we use tryToExecute and split it into separate functions for success and error etc.

MuhammedEdrees (approved):

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/transaction_details/TransactionDetailsScreen.kt
Ahmed7osny1: we cannot use named arguments when calling function type parameters

Malak187 (approved):

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/transaction_details/TransactionDetailsScreen.kt
MuhammedEdrees: Just noticed that in onCaptureError we are treating snackbar as an effect not a ui state. This is inconsistent with the other viewmodels and should be changed.

Malak187: I noticed it too but i thought this can be resolved in another PR for refactoring in general. What do you think? @MuhammedEdrees

Ahmed7osny1: Agree with you @Malak187

MuhammedEdrees: I think it's a small edit not worth a separate PR but as you like.

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/export/ExportTransactionsViewModel.kt
```diff
@@ -312,48 +313,44 @@ class ExportTransactionsViewModel(
...
+    private suspend fun onDownloadSuccess(filePath: String) {
+        resetDownloadState()
+        showSnackBar(
+            title = getString(Res.string.download_complete),
+            message = getString(Res.string.download_success, filePath),
+            isSuccess = true
+        )
+    }
+
+    private suspend fun onDownloadFailure(error: ErrorState) {
```
hagarabobakr, line 340: Please use handleError () function here to keep the file consistent.

hagarabobakr (approved):

Farah315 (approved):
great work 👏🏻

EL-MANCY (approved):

PR #398, opened 2025-10-15
feature/DKN-65-dukan-main

mostafa-n3ma:
Scop:
main screen implementation 

why wee need this :
we need to display dukans categories ,best dukans around and editor picks dukans list 
with navigating to categories screen and dukans under category  screen

https://github.com/user-attachments/assets/ea596b23-cafc-43f7-9788-10a24d227ec6


dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/dukanCategories/DukanCategoriesScreen.kt
```diff
@@ -0,0 +1,179 @@
...
+                        .fillMaxWidth()
+                        .clip(RoundedCornerShape(Theme.radius.md))
+                        .clickable(
+                            onClick = interactionListener::onDismissSnackBar,
+                            indication = null,
+                            interactionSource = null
+                        ),
+                    onDismiss = interactionListener::onDismissSnackBar,
+                    snackBarUiState = snackBarState
+                )
```
YoussefmSaber, line 97: use the snackbar the is provided from the design system, the onClick on the snack bar is not our responsibility

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/dukanCategories/DukanCategoriesScreen.kt
```diff
@@ -0,0 +1,179 @@
...
+            CategoryCard(
+                title = category.name,
+                imageUrl = category.imageUrl,
+                onClick = { onCategoryClick(category.name, category.id) },
+            )
+        }
+    }
+}
+
+private val categoryItemSize = 76.dp
```
YoussefmSaber, line 163: if this is only used inside the categoriesList function, then declare it inside it, don't leave it out here

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/manageDukan/component/ProductsList.kt
```diff
@@ -65,7 +65,7 @@ private fun ProductsLayoutPreview() {
             fakeProducts(),
             pager = Pager(
                 config = PagingConfig(),
-                pagingSourceFactory = { FakeProductPagingSource() }
+                pagingSourceFactory = { FakeProductPagingSource }
```
YoussefmSaber, line 68: shouldn't you send a lambda here

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/util/stubPreviews/BestNearestDukanPagingSource.kt
```diff
@@ -0,0 +1,44 @@
+package net.thechance.mena.dukan.presentation.util.stubPreviews
+
+import net.thechance.mena.dukan.domain.util.PagedResult
+import net.thechance.mena.dukan.presentation.util.pagination.base.BasePagingSource
+import net.thechance.mena.dukan.presentation.viewModel.mainScreen.MainScreenUiState
+
+object BestNearestDukanPagingSource :
```
YoussefmSaber: if this is a fake paging source, add to it a fake name to not be missed up with the real one

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/viewModel/mainScreen/MainViewModel.kt
```diff
@@ -15,9 +22,111 @@ class MainViewModel(
...
+            copy(
+                bestNearestDukans = dukans,
+                bestNearestDukanState = loadedBestNearestDukans
+            )
+        }
+    }
+
+    private fun getCategories() {
+        tryToExecute(
+            onStart = { updateState { copy(dukanState = MainScreenUiState.DukanState(status = DukanStatusUi.Loading)) } },
```
YoussefmSaber: extract the content to a private function

YoussefmSaber (commented):
Amazing work, but there is some modifications need to be done

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/dukanCategories/DukanCategoriesScreen.kt
ahmedNaser7: we use snackbar component of dukan feature  , the snackbar now is clickable in design as we discuss in meeting , so we only use clickable of modifier and if they change it later we will update

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/dukanCategories/DukanCategoriesScreen.kt
ahmedNaser7: its only for readability and i think The best practice is to keep it outside composables at the file level with private visibility to avoid the recreation of object on every recomposition inside composable function

mostafa-n3ma: it private and in the same file it is used in so keeping it outside is preferred I think it is not depend on recomposition and can be reused in other composables like previews or other

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/manageDukan/component/ProductsList.kt
mostafa-n3ma: In this case, FakeProductPagingSource is used only for the preview/testing setup, so we don’t need to instantiate a new source on each call like the real paging behavior.
Using the singleton object directly here keeps the preview simple and avoids unnecessary instantiation logic, since no actual paging or refreshing happens in this context.

haideraljawasim (approved):

ElsayedMagdy122 (approved):

YoussefmSaber (approved):

ahmedNaser7 (approved):

Ahmed-AlMancy (approved):

PR #466, opened 2025-10-20
Refactor CHT 78 send messages

dev-bilal-azzam:
- Use strategy pattern to abstract sending messages for TextMessage and ImagesMessage
- Use factory pattern to encpsulate `MessageSender` creation based on the `MessageContent`


core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/utils/tryNetworkCall.kt
```diff
@@ -0,0 +1,69 @@
...
+): T? {
+    return runCatchingWithException(defaultException) {
+        retry(maxAttempts) {
+            val response = call()
+            response.getSuccessBodyOrThrow(bodyType)
+        }
+    }
+}
+
+private suspend fun <T> retry(
```
ZainabAlnajjar, line 30: It would be better to leave changing on this file to another pr, since it need more refactorinh like removing this retry function and the maxAttempts

ZainabAlnajjar (approved):
Nice work, Bilal

Mohamed-Elhanafy (approved):

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/utils/tryNetworkCall.kt
dev-bilal-azzam: this is just temporal because I needed it, So that I didn't refactor it.

AdelTamer35 (approved):

ZainabAlnajjar (approved):

PR #502, opened 2025-10-21
Refactor/CHT 78 remove chat effector

ZainabAlnajjar:
# What have been done
- remove chatEffector to decuple dependency between navigation library and BaseViewModel
- add `LocalNavController` to handle navegation between screens
- add  `LocalSnackBarHostController`  to control the snackBar from any screen without interrupting its animation when navigating to another screen
- refactor related tests
- fix bug with mockery plugin setup not provided properly and failing tests


core_chat_presentation/src/commonMain/kotlin/net/thechance/mena/core_chat/presentation/screen/contacts/ContactsScreen.kt
```diff
@@ -123,4 +135,54 @@ private fun ContactsContent(
...
+    val navController = LocalNavController.current
+
+    val stateFlow = navController.currentBackStackEntry
+        ?.savedStateHandle
+        ?.getStateFlow(IS_SYNC_SUCCESS, false)
+
+    val isSyncedState = stateFlow?.collectAsState(initial = false)
+    val isSynced = isSyncedState?.value == true
+
+    LaunchedEffect(isSynced) {
```
dev-bilal-azzam: I think receiving and setting savedStateHandle doesn't belong to `EffectHandler`

core_chat_presentation/src/commonMain/kotlin/net/thechance/mena/core_chat/presentation/shared/BaseViewModel.kt
```diff
@@ -25,45 +26,27 @@ import kotlinx.coroutines.flow.onEmpty
...
-        viewModelScope.launch { effector.setNavigationArgs(*arguments) }
-
-    protected fun popBackStack(vararg arguments: Pair<String, Any>) =
-        viewModelScope.launch { effector.popBackStack(*arguments) }
+    protected fun updateState(updater: (S) -> S) = _state.update(updater)
 
-    protected fun showSnackBar(snackBarData: SnackBarData) {
-        viewModelScope.launch {
-            effector.showSnackBar(snackBarData)
+    fun emitEffect(effect: E) {
```
dev-bilal-azzam, line 50: I think it is necessary to prevent emitting effects frequently, because this might cause ui bugs like redundunt navigation or triggering showing multiple snackbars at the same time.
we might need throttleFirst bahviour to ensure emitting just one effect in a specified time range.

dev-bilal-azzam: something like
```
    private var lastEffect: E? = null
    private var lastEffectTime: Long = 0L


    fun emitEffect(effect: E) {
        viewModelScope.launch(defaultDispatcher) {
            val now = Clock.System.now().toEpochMilliseconds()
            if (effect != lastEffect ||  now - lastEffectTime > 500) {
                _effect.emit(effect)
                lastEffect = effect
                lastEffectTime = now
            }
        }
    }


```

omer1998 (approved):
good job zainab

Radwa85 (approved):

Radwa85 (approved):

raouf-kamel (approved):

Al-Qassim (approved):
just make sure you are calling `dismissSnackBar` after 3 seconds of `showSnackBar`

PR #513, opened 2025-10-22
Refactor/cht 78 camera manager

ZainabAlnajjar:
# What have been done
- refactor `CameraManager `to use `FileKit`
- move image compression logic to viewModel
- ensure code is working on both android and ios


Al-Qassim (approved):
take note that, the ios and the android implementation are exactly the same, so try to not use `actual` and `expect`, just put the code directly in the common main

ZainabAlnajjar:
@Al-Qassim I noticed this but it is an issue with FileKit library, currently, it doesn't support running camera code in commonMain directly, also the returned file type from each module (ios, android) is different
hopefully they would fix it in future updates

omer1998 (approved):
good job zainab

AdelTamer35 (approved):

Radwa85 (approved):

Radwa85 (approved):

raouf-kamel (approved):

PR #515, opened 2025-10-22
Refactor/TRN 79 separate observe uploading

ahmedmahgoub0:
## Description
- Separate the reel upload request from upload progress observing
- Remove `UploadReelSuccess` and make it only to `UploadProgress`


trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/repository/ReelsRepositoryImpl.kt
```diff
@@ -50,6 +50,9 @@ internal class ReelsRepositoryImpl(
     @Provided private val videoFileHandler: VideoFileHandler
 ) : ReelsRepository {
 
+    private val observableUploadingFlow: MutableStateFlow<UploadReelProgress> =
```
ahmeddsamii: what about using SharedFlow here? it's hot flow and in the same time it doesn't require this initial value, what do you think?

mohamedd-hassan (approved):

yusufnasserdev (approved):

ahmeddsamii (approved):

iZyadNasser (approved):
LGTM

PR #518, opened 2025-10-22
Fix TRN-78 client separation

HendSayed25:
- Separate HttpClient instances for default and upload operations to improve network performance and configuration flexibility.
- Refactor Reel Repository with new client
- Made category Id non-nullable 
- Update Reel Repository Test and category Mapper


trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/client/NetworkClient.kt
```diff
@@ -27,61 +29,82 @@ class NetworkClient : KoinComponent {
...
-                url(baseUrl)
-            }
+            baseConfig(this)
 
             install(Logging) {
-                level = LogLevel.HEADERS
-                filter { request ->
-                    request.body !is MultiPartFormDataContent
-                }
+                level = LogLevel.BODY
```
ahmeddsamii: I think this should be .ALL

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/dto/CategoryDto.kt
```diff
@@ -6,7 +6,7 @@ import kotlinx.serialization.Serializable
 @Serializable
 internal data class CategoryDto(
     @SerialName("id")
-    val id: String? = null,
```
ahmeddsamii, line 9: why?

ahmeddsamii: You can in mapper to do id.orEmpty()

mohamedd-hassan (approved):

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/client/NetworkClient.kt
```diff
@@ -27,61 +29,82 @@ class NetworkClient : KoinComponent {
...
-                            refreshToken = authorizationService.refreshToken()
-                        )
+    @Single
+    fun provideUploadHttpClient(): HttpClient {
+        return HttpClient(engine = getHttpEngine()) {
+            baseConfig(this)
+
+            install(Logging) {
+                level = LogLevel.HEADERS
+                filter { request -> request.body !is MultiPartFormDataContent }
```
ahmedmahgoub0: I think we don't need this filter if we already log HEADERS,
@ahmeddsamii ?

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/repository/ReelsRepositoryImpl.kt
```diff
@@ -131,7 +133,7 @@ internal class ReelsRepositoryImpl(
 
     override suspend fun uploadReelThumbnail(reelId: String, thumbnail: ByteArray) {
         safeApiCall<Unit> {
-            networkClient.put(urlString = "$THUMBNAIL_ENDPOINT/${reelId}") {
+            uploadClient.put(urlString = "$THUMBNAIL_ENDPOINT/${reelId}") {
```
ahmedmahgoub0, line 136: @ahmeddsamii @iZyadNasser 
just asking if we need same upload client for thumbnail 
image size is less that 1MB so I think using default client here is better?

ahmedmahgoub0 (approved):
just some points need to be discussed
Good work👏👏

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/client/NetworkClient.kt
HendSayed25: You are right this headers not contains the body .. But I added it as same before not changed any thing

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/dto/CategoryDto.kt
HendSayed25: I did this according to update type of id to UUID as said in code review and remove null default value

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/repository/ReelsRepositoryImpl.kt
ahmeddsamii: It should be replaced with the upload one

ahmedmahgoub0: can I ask why? @ahmeddsamii

ahmeddsamii: We created the other client for multipart file, so even if it's video or image we should use it

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/client/NetworkClient.kt
ahmedmahgoub0: yes we made this as a work around 
but now we have 2 clients so no need to filter body if log level is headers

ahmeddsamii (approved):

ahmeddsamii (approved):

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/client/NetworkClient.kt
```diff
@@ -27,61 +28,81 @@ class NetworkClient : KoinComponent {
...
             install(Logging) {
-                level = LogLevel.HEADERS
-                filter { request ->
-                    request.body !is MultiPartFormDataContent
-                }
+                level = LogLevel.ALL
                 logger = object : Logger {
                     override fun log(message: String) {
-                        println("Http client: $message")
+                        println("Default Http Client: $message")
```
iZyadNasser: Why not use Kermit logger?

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/client/NetworkClient.kt
```diff
@@ -27,61 +28,81 @@ class NetworkClient : KoinComponent {
...
+                    ignoreUnknownKeys = true
+                }
+            )
+        }
     }
 
     private companion object {
-        const val TIME_OUT_INTERVAL_MILLI = 15_000L
+        const val TIME_OUT_INTERVAL_MILLI = 60_000L
         const val BASE_URL = "baseUrl"
```
iZyadNasser: I know this wasn't part of your task, but since you are working on this file anyway, maybe change this const name to (BASE_URL_KEY) or something like that

trends_data/src/commonMain/kotlin/net/thechance/mena/trends/data/mapper/CategoryMapper.kt
```diff
@@ -5,9 +5,8 @@ import net.thechance.mena.trends.data.util.orFalse
 import net.thechance.mena.trends.domain.entity.Category
 
 internal fun CategoryDto.toEntity(): Category? {
-    if (id == null) return null
     return Category(
-        id = id,
+        id = id.orEmpty(),
```
iZyadNasser: I don't think this solves the problem Bareq was talking about, if we got an id that is null, it will act normally as if there was nothing wrong, until we try to do an operation on that id (which we might not do), then it will crash

I think it's better to make the id non-nullable so that if it's null, an error will happen and we will know right away that something is wrong with the backend

ahmeddsamii: if the id is null or not this will not solve the problem if the backend give it to us with null value and this case will NEVER happens

@iZyadNasser

iZyadNasser (approved):
Great Job

ahmeddsamii (approved):

PR #526, opened 2025-10-23
feat: Introduce safe invocation in `tryToExecute`

hagarabobakr:


wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/base/BaseViewModel.kt
MuhammedEdrees, line 1: Why do we need this change

Joseph-Sameh-0, line 1: why you catching CancellationException ?
I think it's better to use coroutine exception handler

PR #532, opened 2025-10-24
Refactor/TRN 79 split error state

fatimafaisalghazi:
Make ErrorState a base error state.
Split UploadErrorState from it.
Make mapExceptionToErrorState a `protected open suspend fun mapExceptionToErrorState`,
to allow the ViewModel of each screen to have a separate error state and override it.
rename : ioDispatcher to defaultDispatcher


trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/upload_reel/UploadReelViewModel.kt
```diff
@@ -76,7 +81,8 @@ internal class UploadReelViewModel(
     }
 
     private fun onValidationError(errorState: ErrorState) {
-        updateState { copy(errorState = errorState) }
+        val uploadReelError = mapToUploadReelError(errorState)
+        updateState { copy(errorState = uploadReelError) }
```
HendSayed25: what's your opinion to add this line in update direct -> mapToUploadReelError(errorState) 
I think not need anew variable or function Like this :
```suggestion
        updateState { copy(errorState = mapToUploadReelError(errorState)) }
```

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/upload_reel/UploadReelViewModel.kt
```diff
@@ -196,12 +205,41 @@ internal class UploadReelViewModel(
         tryToExecute(
             block = { state.value.reelId?.let { reelsRepository.deleteReelById(id = it) } },
             onSuccess = { updateState { UploadReelScreenState() } },
-            onError = { errorState -> updateState { copy(errorState = errorState) } },
+            onError = { errorState -> onClickDeleteVideoError(errorState) },
```
HendSayed25: I think not need to this function only become : 
```suggestion
            onError = { errorState -> mapToUploadReelError(errorState) },
```

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/upload_reel/UploadReelViewModel.kt
```diff
@@ -196,12 +205,41 @@ internal class UploadReelViewModel(
...
+        val errorState = when (throwable) {
+            is NoInternetException -> ErrorState.NoInternet
+            is MaxFileSizeExceededException -> UploadReelErrorState.FileTooLarge
+            is MaxFileDurationExceededException -> UploadReelErrorState.DurationTooLarge
+            else -> ErrorState.RequestFailed(throwable.message)
+        }.also { errorState -> Logger.e(TAG) { errorState.toString() } }
+        onError(errorState)
+    }
+
+    private companion object {
```
HendSayed25, line 236: I suggest add the companion object in the end of class after all functions

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/upload_reel/UploadReelViewModel.kt
```diff
@@ -196,12 +205,41 @@ internal class UploadReelViewModel(
...
+    private fun onClickDeleteVideoError(errorState: ErrorState) {
+        val uploadReelError = mapToUploadReelError(errorState)
+        updateState { copy(errorState = uploadReelError) }
+    }
+
     override fun onClickRetryUpload() {
         uploadTrend()
     }
+
+    override suspend fun mapExceptionToErrorState(
```
HendSayed25, line 219: why become override this function in all view models and we already has mapToUploadReelError function ?

fatimafaisalghazi: this not override in all view model ,
it just in here ,
becase in the base view model we map to Error state ,
in here we map to Upload Error state ,
Error state is the parent and used in all screen ,
upload it is onley to upload screen

fatimafaisalghazi: mapToUploadReelError 
since i have to pass Error State to Try To Excete , try to collect ,
so i map to Upload to avoid type mis match

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/upload_reel/UploadReelViewModel.kt
fatimafaisalghazi:
```
  private fun mapToUploadReelError(errorState: ErrorState): UploadReelErrorState {
        return when (errorState) {
            is UploadReelErrorState -> errorState
            else -> UploadReelErrorState.RequestTimeout
        }
    }
```
    
    it is a mapper dosenot have a update state
    
```
       private fun onClickDeleteVideoError(errorState: ErrorState) {
        val uploadReelError = mapToUploadReelError(errorState)
        updateState { copy(errorState = uploadReelError) }
    }
```

HendSayed25: sorry but I don't understand you

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/upload_reel/UploadReelViewModel.kt
HendSayed25: I think if we collect all error in base view model and each screen will map the error message to what it want using the map error function like you made , I think this do the same thing , what about your opinion ?

fatimafaisalghazi: Think about having an `Upload Error State` and then for example  we will have `User Reels Error State`
. This will make us modify the base ViewModel to accept the `User Reels Error State` , 
which must take the generic type, and I don’t think it is okay to modify it.

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/upload_reel/UploadReelViewModel.kt
fatimafaisalghazi: i its not okey to do your suggeston becase these function  dose not do the same thing 
one of theme have to updateState

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/base/ErrorState.kt
```diff
@@ -2,8 +2,11 @@ package net.thechance.mena.trends.presentation.shared.base
...
-    object RequestTimeout : ErrorState()
     data class RequestFailed(val message: String? = "Request failed") : ErrorState()
-    object FileTooLarge : ErrorState()
-    object DurationTooLarge : ErrorState()
+}
+
+sealed class UploadReelErrorState: ErrorState() {
+    object FileTooLarge : UploadReelErrorState()
+    object DurationTooLarge : UploadReelErrorState()
+    object RequestTimeout: UploadReelErrorState()
```
ahmeddsamii: RequestTimeout is shared between all error states, i think you should keep it in the error state

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/upload_reel/UploadReelViewModel.kt
```diff
@@ -196,12 +201,40 @@ internal class UploadReelViewModel(
...
+    private fun onClickDeleteVideoError(errorState: ErrorState) {
+        updateState { copy(errorState = mapToUploadReelError(errorState)) }
+    }
+
     override fun onClickRetryUpload() {
         uploadTrend()
     }
+
+    override suspend fun mapExceptionToErrorState(
+        throwable: Throwable,
```
ahmeddsamii, line 220: We don't need this function here, and we can do that in the map of the baseViewModel like this 


private suspend fun mapExceptionToErrorState(
throwable: Throwable,
onError: suspend (ErrorState) -> Unit,
) {
logError(throwable)
val message = throwable.message
when (throwable) {
is NoInternetException -> ErrorState.NoInternet
is MaxFileSizeExceededException -> UploadReelErrorState.FileTooLarge
is MaxFileDurationExceededException -> UploadReelErrorState.DurationTooLarge
else -> ErrorState.RequestFailed(message).also { logError(throwable) }
}.also { errorState ->
Logger.e(LOG_TAG){errorState.toString()}
}.let { onError(it) }
}

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/base/BaseViewModel.kt
```diff
@@ -104,16 +100,14 @@ internal abstract class BaseViewModel<State, Effect>(
         }
     }
 
-    private suspend fun mapExceptionToErrorState(
+    protected open suspend fun mapExceptionToErrorState(
```
ahmeddsamii, line 103: return this to be prive and implement it like this 


private suspend fun mapExceptionToErrorState(
throwable: Throwable,
onError: suspend (ErrorState) -> Unit,
) {
logError(throwable)
val message = throwable.message
when (throwable) {
is NoInternetException -> ErrorState.NoInternet
is MaxFileSizeExceededException -> UploadReelErrorState.FileTooLarge
is MaxFileDurationExceededException -> UploadReelErrorState.DurationTooLarge
else -> ErrorState.RequestFailed(message).also { logError(throwable) }
}.also { errorState ->
Logger.e(LOG_TAG){errorState.toString()}
}.let { onError(it) }
}

ahmeddsamii (commented):
I've few notes, kindly check them and thank you for your effort

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/upload_reel/UploadReelViewModel.kt
fatimafaisalghazi: The` base ViewModel `is generic, so wouldn’t it be a problem if we did this?
Like, if we put the `UploadReelErrorStat`e there
 — imagine we’ll later have another separate error state that would force us to modify it and add the new one.

What’s your opinion?

Would we  keep it in the base and modify it whenever we have a new error state?
Or should we make it open and override it when we need a new error state?

Thank you Ahmed for your note!

ahmeddsamii: i think it's okay to keep it in the base viewModel and do mapping there, we only want to separate different error states away but the mapping i think it's okay to keep it in the base viewModel and implement a helper function like what you did

ahmeddsamii: what's your opinion? @iZyadNasser

iZyadNasser: I agree with fatima, I think if we add this mapping to the base view model, it will violate the OCP, whenever we have new screen with its error state, we will have to modify the base view model, which should be generic. However, I think only the screen specific error state should be mapped here, the base error state should still be mapped in base view model, maybe an approach like this:

```
override suspend fun mapExceptionToErrorState(
        throwable: Throwable,
        onError: suspend (ErrorState) -> Unit
    ) {
        when (throwable) {
            is MaxFileSizeExceededException -> onError(UploadReelErrorState.FileTooLarge)
            is MaxFileDurationExceededException -> onError(UploadReelErrorState.DurationTooLarge)
            else -> {
                super.mapExceptionToErrorState(throwable, onError)
            }
        }.also { errorState -> Logger.e(TAG) { errorState.toString() } }
    }
```

iZyadNasser: This is also a better approach, because if we decided to add more error states in the base error state in the future, we won't have to add its handling to all view models that overrides this function, (DRY)

iZyadNasser: I still think there is actually a better approach than all this but I'm not sure, this is the best approach I can think of for now
"شاكك إن فيه طريقة أشيك بس معنديش دليل"

HendSayed25 (approved):
Good work Fatima

ahmedmahgoub0 (approved):
Good work👏

ahmeddsamii (commented):
put this line in onValidationError function  
sendEffect(UploadReelScreenEffect.ShowErrorSnackbar(errorState = errorState))

to be like this

```
private fun onValidationError(errorState: ErrorState) {
        updateState { copy(errorState = errorState) }
        sendEffect(UploadReelScreenEffect.ShowErrorSnackbar(errorState = errorState))
    }
```

and fix tests

ahmeddsamii (approved):

PR #533, opened 2025-10-24
Refactor/idn 110 save profile image in cache

ShroukMohamed16:
- This PR refactors the navigation flow to stop passing raw `ImageBitmap `objects as navigation arguments.

- The new implementation saves the ImageBitmap to a shared cache (`ImageCacheManager`) on the source screen and passes a lightweight, unique string key to the destination screen. The destination screen then uses this key to retrieve the bitmap from the cache.

- refactor `imageCropperViewModel `Tests


AhmadKharfan (approved):

identity_presentation/src/commonMain/kotlin/net/thechance/mena/identity/presentation/utils/ImageCacheManager.kt
```diff
@@ -0,0 +1,9 @@
+package net.thechance.mena.identity.presentation.utils
+
+import androidx.compose.ui.graphics.ImageBitmap
+
+interface ImageCacheManager {
```
nabilBouzineDev: Well Done @ShroukMohamed16 👏 

But, I think this abstraction layer is unnecessary and we can use the expect class directly like the case of the `PermissionManager` class

shahad-97 (approved):

ThoriaHamdy99 (approved):

identity_presentation/src/androidMain/kotlin/net/thechance/mena/identity/presentation/utils/ImageCacheManagerImpl.android.kt
```diff
@@ -0,0 +1,20 @@
+package net.thechance.mena.identity.presentation.utils
+
+import androidx.compose.ui.graphics.ImageBitmap
+
+actual class ImageCacheManagerImpl : ImageCacheManager {
+    val cache = mutableMapOf<String, ImageBitmap>()
```
poula22: we need to `cache it into temp file` not a map. The reason of caching is we don't need pass this large object direct from memory from class to another so we save into temp file that will be deleted automatically.

identity_presentation/src/androidMain/kotlin/net/thechance/mena/identity/presentation/utils/ImageCacheManagerImpl.android.kt
```diff
@@ -0,0 +1,20 @@
+package net.thechance.mena.identity.presentation.utils
+
+import androidx.compose.ui.graphics.ImageBitmap
+
+actual class ImageCacheManagerImpl : ImageCacheManager {
```
poula22: This class is a `datasource` so we to move it to `data module`.

I think these should be functions in `repository` to `saveTempProfileImage` and `getTempProfileImage` and then we access it from `viewModel`. if we need it to be more general we can create `BitmapRepository` with these 2 functions without Profile word in names.

identity_presentation/src/iosMain/kotlin/net/thechance/mena/identity/presentation/utils/ImageCacheManagerImpl.ios.kt
```diff
@@ -0,0 +1,32 @@
+package net.thechance.mena.identity.presentation.utils
+
+import androidx.compose.ui.graphics.ImageBitmap
+import kotlinx.cinterop.BetaInteropApi
+import platform.Foundation.NSCache
+import platform.Foundation.NSString
+import platform.Foundation.create
+
+@OptIn(BetaInteropApi::class)
+actual class ImageCacheManagerImpl : ImageCacheManager {
```
poula22: As I mentioned before this should be moved to data module

poula22 (changes requested):

identity_presentation/src/iosMain/kotlin/net/thechance/mena/identity/presentation/utils/ImageCacheManagerImpl.ios.kt
```diff
@@ -0,0 +1,32 @@
...
+    private val cache = NSCache()
```
poula22: Perfect to use here for IOS👏

identity_presentation/src/androidMain/kotlin/net/thechance/mena/identity/presentation/utils/ImageCacheManagerImpl.android.kt
nabilBouzineDev: That's right, but I suggest instead of making that manually. We can use `LRU` cache which removes automatically unused items using the garbage collector

https://developer.android.com/reference/android/util/LruCache

poula22 (approved):

nabilBouzineDev (approved):

ABDULLAHHG (approved):

AhmadKharfan (approved):

PR #537, opened 2025-10-25
Fix/TRN-78-update home screen state handling

ahmeddsamii:
The home screen's state handling is updated to rely on the `PagingItems` load state instead of the ViewModel's `isLoading` state. This change improves the accuracy of loading, empty, and error state representations.

### DIFF
--- a/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/home/HomeScreen.kt
+++ b/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/home/HomeScreen.kt
@@ -138,7 +138,7 @@ private fun HomeScreenContent(
 
             AnimatedVisibility(
                 modifier = Modifier.align(Alignment.BottomEnd),
-                visible = reels.loadState.refresh !is LoadState.Loading,
+                visible = shouldShowEmptyState && !hasNetworkError,
                 content = { AddTrendFAB(onClickFab = { listener.onClickAddReel() }) }
             )
         }
--- a/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/home/HomeScreen.kt
+++ b/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/home/HomeScreen.kt
@@ -107,7 +107,7 @@ private fun HomeScreenContent(
 
         val shouldShowEmptyState = reels.itemSnapshotList.isEmpty() &&
                 reels.loadState.refresh is LoadState.NotLoading &&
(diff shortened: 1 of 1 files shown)


mohamedd-hassan (approved):

Shahd-Hatem (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/home/HomeScreen.kt
```diff
@@ -100,18 +101,22 @@ private fun HomeScreenContent(
         }
     ) {
         val reels = state.reels.collectAsLazyPagingItems()
+
+        val hasNetworkError = reels.loadState.refresh.toErrorState() == ErrorState.NoInternet
+                && reels.itemSnapshotList.isEmpty()
```
fatimafaisalghazi, line 106: What happens if the internet connection goes off in the middle of watching a reels?
Will the condition `&& reels.itemSnapshotList.isEmpty()` have any effect or not?

ahmeddsamii: if we already load some posts so the offline screen will not be shown

fatimafaisalghazi (approved):
great !

yusufnasserdev (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/home/HomeScreen.kt
```diff
@@ -121,7 +126,7 @@ private fun HomeScreenContent(
             )
 
             AnimatedVisibility(
-                visible = reels.itemSnapshotList.isNotEmpty() && state.isLoading.not(),
+                visible = reels.itemSnapshotList.isNotEmpty() && reels.loadState.refresh !is LoadState.Loading,
```
HendSayed25, line 129: reels.loadState.refresh !is LoadState.Loading 
 I think this line will return true when not loading or error , right or?

ahmeddsamii: when no loading or empty or network error, the content will be shown and if we already have cached data then it will be shown

ahmeddsamii: when no loading or empty or network error, the content will be shown and if we already have cached data then it will be shown

HendSayed25 (approved):
Nice improvement !

iZyadNasser (approved):

PR #571, opened 2025-10-29
Fix/TRN 86 upload feature

ahmedmahgoub0:
## Description
- Change android client engine from `CIO` to `OkHttp` 
- Remove `suspend` when reading file 
- Remove `getFileUri` extension and use `path` extension from library
- Fix `NextButton` behavior in `AddDescriptionScreen` when keyboard is active

### DIFF
--- a/trends_data/build.gradle.kts
+++ b/trends_data/build.gradle.kts
@@ -22,7 +22,7 @@ kotlin {
 
     sourceSets {
         androidMain.dependencies {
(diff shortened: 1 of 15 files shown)


ahmeddsamii (approved):
Good job, mahgoub

iZyadNasser (approved):

mohamedd-hassan (approved):

PR #587, opened 2025-10-30
Feature/idn 132 implement domain layer for register

ABDULLAHHG:
### implement domain layer for register


identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/repository/RegistrationRepository.kt
```diff
@@ -0,0 +1,9 @@
+package net.thechance.mena.identity.domain.repository
+
+import net.thechance.mena.identity.domain.entity.User
+
+interface RegistrationRepository {
+    fun createUser(user: User, password: String)
+    fun uploadImage(imageByteArray: ByteArray)
+    fun usernameExists(username: String): Boolean
```
AhmadKharfan:
```suggestion
    fun isUsernameTaken(username: String): Boolean
```

AhmadKharfan (commented):
Good job but what about the OTP and create password?

AhmadKharfan (approved):

PR #592, opened 2025-10-30
fix:UI in Dukan screens

AbdelrhmanAyman20470:
- handle max lines 
- remove horizontal padding in each item in create dukan screen, use content padding in lazy column 
- fix ripple on product edit icon 
- fix icon padding with category selection

### DIFF
--- a/dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/component/chip/SelectionRow.kt
+++ b/dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/component/chip/SelectionRow.kt
@@ -22,12 +22,11 @@ fun SelectionRow(
     isItemSelected: (CreateDukanUiState.DukanCategoryUiState) -> Boolean,
     onItemClicked: (CreateDukanUiState.DukanCategoryUiState) -> Boolean,
     onItemEnabled: (CreateDukanUiState.DukanCategoryUiState) -> Boolean,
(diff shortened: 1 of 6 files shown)


Ahmed-AlMancy (approved):

NadaMansour20 (approved):

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/component/chip/SelectionRow.kt
```diff
@@ -20,10 +22,12 @@ fun SelectionRow(
     isItemSelected: (CreateDukanUiState.DukanCategoryUiState) -> Boolean,
     onItemClicked: (CreateDukanUiState.DukanCategoryUiState) -> Boolean,
     onItemEnabled: (CreateDukanUiState.DukanCategoryUiState) -> Boolean,
+    modifier: Modifier = Modifier
 ) {
     LazyRow(
         contentPadding = PaddingValues(horizontal = Theme.spacing._16),
-        horizontalArrangement = Arrangement.spacedBy(Theme.spacing._8)
+        horizontalArrangement = Arrangement.spacedBy(Theme.spacing._8),
+        modifier = modifier.requiredWidth(getScreenWidth())
```
YoussefmSaber: if you going to use the modifier only here and not control the component from outside, then just call Modifier, no need to add it as optional parameter

YoussefmSaber (approved):

PR #618, opened 2025-11-01
Fix/wlt 109 circular dependency in login

hagarabobakr:
fix circular dependency in login


MennaSrour (approved):
good work Hager

gehadepada (approved):

nourelhodaahmed (approved):

EL-MANCY (approved):

Farah315 (approved):

Farah315 (approved):

Joseph-Sameh-0 (approved):

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/data/remote/client/NetworkClient.kt
```diff
@@ -62,15 +61,16 @@ class NetworkClient(
...
+                        settings.accessToken = newAccessToken
                         BearerTokens(
-                            refreshToken = adminAuthentication.refreshAccessToken(),
-                            accessToken = settings.accessToken
-
+                            accessToken = newAccessToken,
+                            refreshToken = settings.refreshToken
                         )
                     }
                 }
```
Malak187: What about injecting the authenticationService in the NetworkClient and use it in that block like this?

```suggestion
bearer {
                    loadTokens {
                        BearerTokens(
                            accessToken = authorizationService.getAccessToken(),
                            refreshToken = authorizationService.refreshToken()
                        )
                    }
                    refreshTokens {
                        BearerTokens(
                            accessToken = authorizationService.refreshToken(),
                            refreshToken = authorizationService.refreshToken()
                        )
                    }
                }
```

hagarabobakr: Most likely, the same circular dependency will occur if I use this method, which is why I used field injection. Could you take another look at the code and tell me if you have any modifications?

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/data/remote/client/NetworkClient.kt
```diff
@@ -18,21 +17,18 @@ import io.ktor.http.ContentType
...
     @Provided
     @Named("baseUrl")
     private val baseUrl: String,
-    @Provided
-    private val settings: Settings,
-    @Provided
-    private val adminAuthentication: AdminAuthenticationRepository
 ) {
+    private val authenticationService: AuthenticationService
+            by inject(AuthenticationService::class.java)
```
MuhammedEdrees: I think it's better to keep it as constructor param but may be convert it to a function like this:
```
private val authenticationService: () -> AuthenticationService
```
this will solve the circular dependency without making the service hardcoded inside the client which makes it harder to test

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/data/utils/executeApiSafely.kt
```diff
@@ -33,7 +34,7 @@ private inline fun <reified T> handleResponse(response: Response<T>): T {
 
     throw when {
         response.status == HttpStatusCode.Unauthorized ->
-            UnknownNetworkException("Unauthorized: " + parseErrorMessage(response))
+            InvalidCredentialsException("Unauthorized: " + parseErrorMessage(response))
```
MuhammedEdrees: Why do we throw invalid credentials exception if the response status code is unauthorized?

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/data/service/AuthenticationService.kt
MuhammedEdrees, line 1: Why is this class in the domain? I think it belongs to data package

Malak187 (approved):

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/data/remote/client/NetworkClient.kt
```diff
@@ -18,21 +17,18 @@ import io.ktor.http.ContentType
...
     @Provided
     @Named("baseUrl")
     private val baseUrl: String,
-    @Provided
-    private val settings: Settings,
-    @Provided
-    private val adminAuthentication: AdminAuthenticationRepository
-) {
+) : KoinComponent {
+    private val authenticationService: AuthenticationService by lazy { get() }
```
MuhammedEdrees: I think it's better to keep as a constructor param but maybe convert it to a function like this:
```
 private val getAuthenticationService: () -> AuthenticationService
```
this will solve circular dependency while keeping the dependecy provided outside the class itself

Ahmed7osny1 (approved):

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/data/service/AuthenticationService.kt
```diff
@@ -0,0 +1,18 @@
+package net.thechance.mena.admin_panel.domain.service
+
+import net.thechance.mena.admin_panel.domain.repository.AdminAuthenticationRepository
+import org.koin.core.annotation.Single
+
+@Single
+class AuthenticationService(
+    private val adminAuthenticationRepository:
+    AdminAuthenticationRepository
```
Ahmed7osny1: can format this in one line

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/data/service/AuthenticationService.kt
```diff
@@ -0,0 +1,18 @@
...
+) {
+    suspend fun getAccessToken(): String {
+        return adminAuthenticationRepository.getAccessToken()
+    }
```
Ahmed7osny1: can here be like this 
```suggestion
    suspend fun getAccessToken(): String = adminAuthenticationRepository.getAccessToken()
    
```

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/data/service/AuthenticationService.kt
```diff
@@ -0,0 +1,18 @@
...
+
+    suspend fun refreshToken(): String {
+        return adminAuthenticationRepository.refreshAccessToken()
+    }
```
Ahmed7osny1: same here

MuhammedEdrees (approved):
Great work 👏

MuhammedEdrees (approved):

nourelhodaahmed (approved):

nourelhodaahmed (approved):

PR #623, opened 2025-11-01
fix: delete trend failed with configuration changes 

ahmedmahgoub0:
## Description
- fix calling `increaseView` each time recomposition happens by adding condition if trend is deleted or not

### DIFF
--- a/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/user_reel/UserReelViewModel.kt
+++ b/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/user_reel/UserReelViewModel.kt
@@ -84,11 +84,12 @@ internal class UserReelViewModel(
     }
 
     override fun increaseReelView(reelId: String) {
-        tryToExecute(
-            block = { reelsRepository.addReelView(reelId) },
-            onError = { error -> updateState { copy(error = error) } },
-            dispatcher = defaultDispatcher
-        )
+        if(state.value.isReelDeleted == null) {
+            tryToExecute(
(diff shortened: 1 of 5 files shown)


mohamedd-hassan (approved):

HendSayed25 (approved):
Good Job 👏👏

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/user_reel/UserReelViewModel.kt
```diff
@@ -84,11 +85,12 @@ internal class UserReelViewModel(
     }
 
     override fun increaseReelView(reelId: String) {
-        tryToExecute(
-            block = { reelsRepository.addReelView(reelId) },
-            onError = { error -> updateState { copy(error = error) } },
-            dispatcher = defaultDispatcher
-        )
+        if(state.value.isReelDeleted == null) {
+            tryToExecute(
```
iZyadNasser, line 89: we can use the on success to increase the views count in the screen state

iZyadNasser (approved):

ahmeddsamii (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/user_reel/UserReelViewModel.kt
ahmedmahgoub0: but maybe user have watched the video before, api call will return success whether the user watched the video or not, so it will look like a lie.

PR #657, opened 2025-11-03
Fix: TRN-94 Update Categories

yusufnasserdev:
### Navigation Enhancement
- Updated navigation behavior in `UpdateCategoriesScreen` to clear back stack when navigating to Home
- Added `popUpTo` with inclusive flag and `launchSingleTop` for proper navigation flow
- Removed `LaunchedEffect` for category loading

### ViewModel Refactoring
- Renamed `getCategories()` to `initializeCategories()` for better semantics

### DIFF
--- a/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/update_categories/UpdateCategoriesScreen.kt
+++ b/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/update_categories/UpdateCategoriesScreen.kt
@@ -9,7 +9,6 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.unit.dp
@@ -75,13 +74,12 @@ internal fun UpdateCategoriesScreen(
                     )
                 )
 
-                navController.navigate(Route.Home)
+                navController.navigate(Route.Home) {
+                    popUpTo(Route.Home) { inclusive = true }
+                    launchSingleTop = true
+                }
             }
         }
-    }
-
-    LaunchedEffect(Unit) {
-        viewModel.getCategories()
     }
 
     UpdateCategoriesScreenContent(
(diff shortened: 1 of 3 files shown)


trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/update_categories/UpdateCategoriesViewModel.kt
```diff
@@ -21,10 +21,10 @@ internal class UpdateCategoriesViewModel(
 ), UpdateCategoriesInteractionListener {
 
     init {
-        getCategories()
+        initializeCategories()
```
HendSayed25: Why rename this function to initialize?
I think getCategories is suitable for this screen because you will get them to update the selected categories not select them for the first time ..
What's your idea for rename?

yusufnasserdev: I see your point, however, as it's only being done once, as an initialization within the `init` block, `initialize` is the proper naming.

Before it was named `get` as it was being done more than once, so get was more suitable.

ahmeddsamii (approved):

mohamedd-hassan (approved):

iZyadNasser (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/update_categories/UpdateCategoriesScreen.kt
```diff
@@ -75,15 +74,14 @@ internal fun UpdateCategoriesScreen(
                     )
                 )
 
-                navController.navigate(Route.Home)
+                navController.navigate(Route.Home) {
+                    popUpTo(Route.Home) { inclusive = true }
+                    launchSingleTop = true
+                }
```
ahmedmahgoub0, line 80: is there a point to not use `navController.navigateUp()` or `popBackStack` ?

ahmedmahgoub0 (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/update_categories/UpdateCategoriesScreen.kt
yusufnasserdev: The objective is to clear the entire back stack and ensure a clean, _fresh_ instance of **Home** is at the root, which a simple `popBackStack` wouldn't guarantee.

ahmeddsamii (approved):

HendSayed25 (approved):

PR #679, opened 2025-11-05
feature/FTH-113-integrate-identity-address-nearby-mosques

Mostafa-alsaygh:
Integrates `locationService` to fetch the user's active address. This address is then used to retrieve and display nearby mosques on the map.

If the user does not have an active or valid address, they are now navigated to the user addresses screen to select or add one.

Other changes include:
- Adding `identityDomain` and `identityApi` dependencies to the presentation layer.
- Introducing a new `UserAddresses` route.
- Renaming `NavigateToUserLocation` effect to `NavigateToAddressesScreen`.


faith_presentation/src/commonMain/kotlin/net/thechance/mena/faith/presentation/feature/mosque/NearbyMosquesViewModel.kt
```diff
@@ -30,9 +61,7 @@ internal class NearbyMosquesViewModel(
 //        TODO("Not yet implemented")
     }
 
-    override fun onCurrentUserLocationClick() {
-//        TODO("Not yet implemented")
-    }
+    override fun onCurrentUserLocationClick() = handleInvalidAddress()
```
mohannadahmed00: Return it as it was. It is not related to this PR.

faith_presentation/src/commonMain/kotlin/net/thechance/mena/faith/presentation/feature/mosque/NearbyMosquesViewModel.kt
```diff
@@ -141,6 +168,8 @@ internal class NearbyMosquesViewModel(
         // TODO: show snack bar with error message (Res.string.no_mosques_found) to the user
     }
 
+    private fun handleInvalidAddress() = sendEffect(NearbyMosquesEffect.NavigateToAddressesScreen)
```
mohannadahmed00: Remove this function as it is unnecessary; it only sends the effect, which can be directly placed where it is used.

mohannadahmed00 (changes requested):

faith_presentation/src/commonMain/kotlin/net/thechance/mena/faith/presentation/feature/mosque/NearbyMosquesViewModel.kt
Mostafa-alsaygh: because it was two same line i made it in one fun

mohannadahmed00 (approved):

faith_presentation/src/commonMain/kotlin/net/thechance/mena/faith/presentation/feature/mosque/NearbyMosquesViewModel.kt
mohannadahmed00: Not necessary since, as I mentioned, it's just a single line.

mohannadahmed00 (approved):

Mostafa-khaled21 (approved):

Abdulrahman-Ragab-01 (approved):

marwanqashwa (approved):

PR #680, opened 2025-11-05
Feature/fth 130 implement service call for adding mosque

Hasa9nAli:
implement mobile service call for adding mosque


faith_data/src/commonMain/kotlin/net/thechance/mena/faith/data/utils/buildMosqueMultipart.kt
```diff
@@ -0,0 +1,46 @@
...
+
+            image?.let { bytes ->
+                append(
+                    key = "image",
+                    value = buildPacket { writeFully(bytes) },
+                    headers = Headers.build {
+                        append(HttpHeaders.ContentType, ContentType.Image.JPEG.toString())
+                        append(
+                            HttpHeaders.ContentDisposition,
+                            "form-data; name=\"image\"; filename=\"${Uuid.random()}_${Clock.System.now()}.jpg\""
```
mohannadahmed00: I think `Uuid.random()` isn't necessary; `Clock.System.now()` should be enough.

Hasa9nAli: we need uuid to make the unique id not duplicated at same file name 
if we lock on link on google map image for example we will see the link as below for example file when open image mosque in the browser 
jBUyuEzg5kn19LQ1bglG8vfRSwTHBXJOBv9By8fM44wTiZbaVB7hNQii5NjLicpBTQtKEP3SVTXD_yFzwDY3nUncg7yqgpO8BTlV4IkvqlFLlM2lqCAw

mohannadahmed00 (approved):

faith_data/src/commonMain/kotlin/net/thechance/mena/faith/data/utils/buildMosqueMultipart.kt
mohannadahmed00: So `Clock.System.now()` isn't necessary; just using `UUID` is fine.

fatmahgazy (approved):

marwanqashwa (approved):

mohannadahmed00 (approved):

PR #695, opened 2025-11-06
Fix/trn 103 solve crash in ios

ahmeddsamii:
- Replaced the `UserReelSource` enum with `isFromHome` and `isFromManageTrends` boolean flags in `UserReelArgs` and `Route.ReelDetails`.

- Fix tests


yusufnasserdev (approved):
Aside from manual testing, knowing what was causing the crash in the first place, **this looks good to me**.

Good Work!

ahmedmahgoub0 (approved):
👏👏

Shahd-Hatem (approved):

Shahd-Hatem:
Great work ,Ahmed👏

mohamedd-hassan (approved):

HendSayed25 (approved):

iZyadNasser (commented):
Isn't there anyway to use enum and avoid the crash? I honestly don't prefer the two booleans solution, first it may cause confusion in some scenarios, for example what if I passed both as true? Also, if we wanted to add a new screen, we will have to add more boolean arguments which I don't feel is correct.

ahmeddsamii:
> Isn't there anyway to use enum and avoid the crash? I honestly don't prefer the two booleans solution, first it may cause confusion in some scenarios, for example what if I passed both as true? Also, if we wanted to add a new screen, we will have to add more boolean arguments which I don't feel is correct.

They both the same thing, you said what if we pass two true, so in enum class we can pass wrong value so it will be the same and when you add a new Enum so you will add it in the when statement, so both do the same thing and both breaks ocp, and i think no we can't 

didn't we discuss this in the meeting?

iZyadNasser:
> > Isn't there anyway to use enum and avoid the crash? I honestly don't prefer the two booleans solution, first it may cause confusion in some scenarios, for example what if I passed both as true? Also, if we wanted to add a new screen, we will have to add more boolean arguments which I don't feel is correct.
> 
> They both the same thing, you said what if we pass two true, so in enum class we can pass wrong value so it will be the same and when you add a new Enum so you will add it in the when statement, so both do the same thing and both breaks ocp, and i think no we can't
> 
> didn't we discuss this in the meeting?

In enum class, you are required to pass some value of that class at compile time, yes you can pass the wrong one but this normal this like passing true instead of false in a boolean, however what I'm talking about is from a design point of view, when you creat an enum with 2 values, you have only 2 possible inputs to the function, the same as the number of screens you want to represent by the enum, however when you create 2 booleans, we now have 4 possible inputs to the function, 2 of which does not make any sense logically, but we still make them possible, its like you added 2 additional dummy values to the enum when we only need 2 not 4, and imagine we add a new screen, in an enum we will simply add another value, but in this boolean way, if you added a 3rd boolean now we will be up to 8 possible combinations for the function input, while we only need 3, and it will grow exponentially with each new screen

iZyadNasser (approved):

PR #699, opened 2025-11-06
Feature/fth 114 implement calculate distance use case

Hasa9nAli:
Implement a use case that calculates the distance (in kilometers) between two geographical points defined by latitude and longitude coordinates.

### DIFF
--- a/faith_domain/src/commonMain/kotlin/net/thechance/mena/faith/domain/usecase/CalculateDistanceUseCase.kt
+++ b/faith_domain/src/commonMain/kotlin/net/thechance/mena/faith/domain/usecase/CalculateDistanceUseCase.kt
@@ -1,0 +1,44 @@
+package net.thechance.mena.faith.domain.usecase
+
+
+import net.thechance.mena.faith.domain.exception.FaithException
+import kotlin.math.*
+
+
+class CalculateDistanceUseCase {
+
+    operator fun invoke(
+        latitude1: Double,
+        longitude1: Double,
+        latitude2: Double,
+        longitude2: Double
+    ): Double {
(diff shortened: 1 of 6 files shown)


faith_domain/src/commonMain/kotlin/net/thechance/mena/faith/domain/usecase/CalculateDistanceUseCase.kt
```diff
@@ -0,0 +1,44 @@
...
+
+
+import net.thechance.mena.faith.domain.exception.FaithException
+import kotlin.math.*
+
+
+class CalculateDistanceUseCase {
+
+    operator fun invoke(
+        latitude1: Double,
```
mohannadahmed00: We have the `Location` entity in our domain, so using it would be better. What do you think?

mohannadahmed00 (changes requested):

faith_domain/src/commonMain/kotlin/net/thechance/mena/faith/domain/usecase/CalculateDistanceUseCase.kt
Hasa9nAli: great note

Hasa9nAli: done

mohannadahmed00 (approved):

Mostafa-alsaygh (approved):

Mostafa-khaled21 (approved):

PR #712, opened 2025-11-06
Refactor/wlt 122 refactor login screen

nourelhodaahmed:
solved:  

- [x] when no internet it shows unknown error
- [x] there is no maximum length in fields
- [x] move bottom padding to button
- [x] loading ends after error snack bar shows

### DIFF
--- a/admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/login/LoginViewModel.kt
+++ b/admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/login/LoginViewModel.kt
@@ -65,6 +65,7 @@ class LoginViewModel(
     }
 
     private suspend fun onLoginError(errorState: ErrorState) {
+        updateState { it.copy(isLoginButtonLoading = false) }
         showSnackBar(
             title = stringProvider.getString(errorState.getErrorSnackBarTitle()),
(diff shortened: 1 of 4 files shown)


EL-MANCY (approved):

MennaSrour (approved):
good work

Joseph-Sameh-0 (approved):

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/login/LoginViewModel.kt
```diff
@@ -61,6 +65,7 @@ class LoginViewModel(
     }
 
     private suspend fun onLoginError(errorState: ErrorState) {
+        updateState { it.copy(isLoginButtonLoading = false) }
```
MuhammedEdrees, line 68: This should already be happening in onFinish

MuhammedEdrees (approved):

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/login/LoginViewModel.kt
nourelhodaahmed: yes but that make the loading ends after the snack bar shows
we want the loading ends first then the error snack bar shows

Ahmed7osny1 (approved):

PR #763, opened 2025-11-07
fix observing user access token

AimanYosofiUlt:
fix observing user access token


composeApp/src/commonMain/kotlin/net/thechance/mena/appEntryPoint/MainEntryViewModel.kt
```diff
@@ -32,7 +32,9 @@ class MainEntryViewModel(
 
     private fun getUserAccessToken() {
         viewModelScope.launch {
-            _state.update { it.copy(userAccessToken = authorizationService.getAccessToken()) }
+            authorizationService.observeAccessToken().collect { userAccessToken ->
```
iBareq, line 35: Why do I need access token in UI State
Do we need to display it to the user

iBareq (changes requested):

PR #765, opened 2025-11-07
fix/DKN-118-image-picker-open-multiple-times

### LINKED ISSUE #752: Issue: The image picker opens multiple times when the user clicks repeatedly.
Tested on: `develop branch`

https://github.com/user-attachments/assets/71b8d157-0b1d-4484-8b70-9a6436ea2fb3

ahmedNaser7:
- add safeLaunch to handle multiple image picker opening
- fix #752


dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/component/product/productImage/UploadProductImage.kt
```diff
@@ -48,14 +54,24 @@ fun UploadProductImage(
...
 
+    val safeLaunch: () -> Unit = remember {
+        {
+            val now = Clock.System.now().toEpochMilliseconds()
+            if (now - lastLaunchMs.value > 1000L) {
+                lastLaunchMs.value = now
+                filePicker.launch()
+            }
+        }
+    }
```
YoussefmSaber, line 71: search for other way to check if the filePicker.launch is called instead of lastLaunchMs, why is that, what if a device is old and super slow and it takes 3 - 4 seconds to lauch the file picker, then you can launch it 2-3 times, so it's better to check if the filePicker is clicked you add a flag to disable the functionality to relaunch it till the flag is removed

ahmedNaser7: i test it in our worst case api 24 and this case dosnt happen , `PickerResultLauncher` in `fileKit` contains only launch method so you cant do something like that , but i will add flag in case it take more time

ElsayedMagdy122 (approved):

YoussefmSaber (approved):

NadaMansour20 (approved):

PR #772, opened 2025-11-07
fix: remove multiple launching of file picker in upload video screen

iZyadNasser:
### This PR fixes the multiple launching of fire picker when clicking on upload video card in upload screen

- I replaced the immediate call of the launcher in the ui with a call to the interaction listener
- The implementation of the interaction listener function in the view model calls and effect, which is throttled by default in base view model
- Increased the throttle window duration for more strict behavior and to avoid multiple launchs


ahmeddsamii (approved):

ahmedmahgoub0 (approved):

MahmoodTarek (approved):
Great work

mohamedd-hassan (approved):

ahmedmahgoub0 (approved):

PR #798, opened 2025-11-08
fix:empty screen appear before loading data

gehadepada:
fix:empty screen appear before loading data
before:

https://github.com/user-attachments/assets/0007433f-07e5-48fc-9775-b9a4cae3d211

after:

https://github.com/user-attachments/assets/18a367fd-8648-4a4f-b236-16941ea915b6

### DIFF
--- a/admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/users_management/UsersManagementScreen.kt
+++ b/admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/users_management/UsersManagementScreen.kt
@@ -68,9 +68,8 @@ private fun UsersManagementScreenContent(
             )
 
             when {
-                state.isLoading && state.users.isEmpty()-> UsersLoadingIndicator()
-
-                state.users.isEmpty() ->{
+                state.users.isEmpty()-> UsersLoadingIndicator()
+                false ->{
                     if (state.query.isNotEmpty()) {
                         UsersSearchEmptyState()
                     } else{
--- a/admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/users_management/UsersManagementScreen.kt
+++ b/admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/users_management/UsersManagementScreen.kt
@@ -68,8 +68,9 @@ private fun UsersManagementScreenContent(
             )
 
             when {
-                state.users.isEmpty()-> UsersLoadingIndicator()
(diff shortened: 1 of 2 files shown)


EL-MANCY (approved):

MuhammedEdrees (approved):

nourelhodaahmed (approved):

MennaSrour (approved):

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/users_management/UsersManagementScreenState.kt
```diff
@@ -13,7 +13,7 @@ data class UsersManagementScreenState(
     val query: String = "",
     val pageInfo : UserPageInfo = UserPageInfo(),
     val sort: SortState = SortState(),
-    val isLoading: Boolean = false,
+    val isLoading: Boolean = true,
```
Ahmed7osny1: i think problem not here 
check code in screen that condition 

state.isLoading && state.users.isEmpty()-> UsersLoadingIndicator()

                state.users.isEmpty() ->{
                    if (state.query.isNotEmpty()) {
                        UsersSearchEmptyState()
                    } else{
                        UsersSearchEmptyState() /*TODO() change to users empty state*/
                    }
                }

and here need to check that loading true or not because if users is empty that mean fetched data and loading be false

gehadepada: look ahmed when the default value for isLoading is false 
then this condition will be false so it will show empty state screen

gehadepada: I will handle this condition

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/users_management/UsersManagementScreen.kt
```diff
@@ -68,9 +68,8 @@ private fun UsersManagementScreenContent(
             )
 
             when {
-                state.isLoading && state.users.isEmpty()-> UsersLoadingIndicator()
-
-                state.users.isEmpty() ->{
+                state.users.isEmpty()-> UsersLoadingIndicator()
+                false ->{
```
Ahmed7osny1: false for what ?

Ahmed7osny1 (approved):

PR #810, opened 2025-11-09
Refacor/ add error state to main container screen

### LINKED ISSUE #794: Issue: Trending: No error/ empty state handel
https://github.com/user-attachments/assets/6f48add2-d7ad-42ed-9181-9eb3d77bfbd0

ahmeddsamii:
image

### DIFF
--- a/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerScreen.kt
+++ b/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerScreen.kt
@@ -53,14 +53,12 @@ private fun MainContainerScreenContent(
         }
 
         TrendsAnimatedVisibility(state.error is ErrorState.RequestFailed) {
-            SomethingWentWrong(
-                onRetry = { listener.onClickRetry() }
-            )
+            SomethingWentWrong { listener.onClickRetry() }
         }
 
         TrendsAnimatedVisibility(state.error is ErrorState.NoInternet) {
             NoConnection(
-                onRetry = { listener.onClickRetry() }
+                onRetry = listener::onClickRetry
             )
         }
     }
--- a/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerScreen.kt
+++ b/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerScreen.kt
@@ -57,9 +57,7 @@ private fun MainContainerScreenContent(
         }
 
         TrendsAnimatedVisibility(state.error is ErrorState.NoInternet) {
-            NoConnection(
-                onRetry = listener::onClickRetry
-            )
+            NoConnection{ listener.onClickRetry() }
         }
     }
 }
--- a/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerScreen.kt
+++ b/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerScreen.kt
@@ -53,11 +53,11 @@ private fun MainContainerScreenContent(
         }
 
         TrendsAnimatedVisibility(state.error is ErrorState.RequestFailed) {
-            SomethingWentWrong { listener.onClickRetry() }
+            SomethingWentWrong(onRetry = listener::onClickRetry)
         }
 
         TrendsAnimatedVisibility(state.error is ErrorState.NoInternet) {
-            NoConnection{ listener.onClickRetry() }
+            NoConnection(onRetry = listener::onClickRetry)
         }
     }
 }
--- a/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerScreen.kt
+++ b/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerScreen.kt
@@ -3,19 +3,15 @@ package net.thechance.mena.trends.presentation.screen.main_container
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
-import mena.trends_presentation.generated.resources.Res
-import mena.trends_presentation.generated.resources.error_generic
 import net.thechance.mena.designsystem.presentation.component.scaffold.Scaffold
 import net.thechance.mena.trends.presentation.navigation.LocalNavController
 import net.thechance.mena.trends.presentation.navigation.Route
 import net.thechance.mena.trends.presentation.shared.base.ErrorState
 import net.thechance.mena.trends.presentation.shared.component.LoadingProgressBar
+import net.thechance.mena.trends.presentation.shared.component.NoConnection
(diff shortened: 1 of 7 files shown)


ahmedmahgoub0 (commented):
this screen does not have content, it is just for check user interests
so what about adding an error state instead of snack bar
something like no network or something went wrong with a button  to retry ?

ahmeddsamii:
> this screen does not have content, it is just for check user interests so what about adding an error state instead of snack bar something like no network or something went wrong with button to retry ?

We can't add no network because this isn't the error, because the error caused by refresh token which we don't have access to do with it and for retry it already retry by itself 4 times every 15 mins so when the first time an error has occured then if the second call was successful it will update to refresh and then enter to the feature, what do you think?

ahmedmahgoub0:
> We can't add no network because this isn't the error, because the error caused by refresh token which we don't have access to do with it and for retry it already retry by itself 4 times every 15 mins so when the first time an error has occured then if the second call was successful it will update to refresh and then enter to the feature


ok but in the video snackbar looks like a spam without saying what is the problem
is there another solution for better ux?

ahmeddsamii:
> > We can't add no network because this isn't the error, because the error caused by refresh token which we don't have access to do with it and for retry it already retry by itself 4 times every 15 mins so when the first time an error has occured then if the second call was successful it will update to refresh and then enter to the feature
> 
> ok but in the video snackbar looks like a spam without saying what is the problem is there another solution for better ux?

I honestly have no idea what better UX can be, we can talk in daily with the team to figure out

MahmoodTarek (approved):
Great work Ahmed

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/SomethingWentWrong.kt
```diff
@@ -0,0 +1,33 @@
...
+import mena.trends_presentation.generated.resources.something_went_wrong_description
+import mena.trends_presentation.generated.resources.something_went_wrong_title
+import net.thechance.mena.designsystem.presentation.component.button.PrimaryButton
+import org.jetbrains.compose.resources.painterResource
+import org.jetbrains.compose.resources.stringResource
+
+@Composable
+fun SomethingWentWrong(
+    modifier: Modifier = Modifier,
+    onRetry: () -> Unit
```
iZyadNasser, line 18: just a simple note, I prefer putting any required parameter before all optional ones like the modifier

iZyadNasser (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/SomethingWentWrong.kt
ahmeddsamii: I agree with u unless when it's callback function, i prefer to keep it the last parameter

iZyadNasser: why

ahmeddsamii: because we can call it like this SomethingWentWrong { *and here what the onRetry do* } since it's the only callback function

ahmeddsamii: and when we have a function inside another function it better to keep it the last parameter because we can open lambda scope as bareq said in its course of kotlin

ahmedmahgoub0 (approved):
👏👏

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/SomethingWentWrong.kt
iZyadNasser: I know this is best practice in normal functions but in compose, when you have a lambda function that it the last parameter and you use it like that `Function { lambda body }`, it 99% of the time means that this function is a "content" function (another composable) used inside the current composable, and for that reason I don't think it's a best practice to use a callback like this `SomethingWentWrong { and here what the onRetry do }` in compose, because when I read it in a composable screen, I will read it as content not a callback, it's better to pass callbacks as normal named parameter, so I don't keep a function as the last parameter in a composable unless it's another content composable function, otherwise, just handle the lambda parameter like any other parameter, that's my opinion

iZyadNasser: As you can see here, you already are using named parameter instead of `SomethingWentWrong { and here what the onRetry do }` approach in this PR (which is correct)

Screenshot 2025-11-10 102727

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerScreen.kt
```diff
@@ -39,21 +37,31 @@ internal fun MainContainerScreen(
...
+        TrendsAnimatedVisibility(
+            visible = state.isCategoriesAlreadySelectedByUser == null
         ) {
-            DotsProgressIndicator()
+            LoadingProgressBar()
+        }
+
+        TrendsAnimatedVisibility(state.error is ErrorState.RequestFailed) {
+            SomethingWentWrong(
+                onRetry = { listener.onClickRetry() }
```
HendSayed25: Isuggest use it as we use in other files.

```suggestion
                onRetry =  listener::onClickRetry
```

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/main_container/MainContainerScreen.kt
```diff
@@ -39,21 +37,31 @@ internal fun MainContainerScreen(
...
+            )
+        }
+
+        TrendsAnimatedVisibility(state.error is ErrorState.NoInternet) {
+            NoConnection(
+                onRetry = { listener.onClickRetry() }
```
HendSayed25: Here also

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/shared/component/SomethingWentWrong.kt
ahmeddsamii: I got your point but all states in our project like this, and we can simply go to function to know its signature, if you go to any state Compose in our feature it will be like this 

image
image

i followed the same structure

ahmeddsamii: and i edited how i call the function because i forgot to push it

iZyadNasser: ok, you can keep the callback as the last parameter for consistency if you like it that way, but don't use trailing lambda, it really messes with compose readability a lot on the long term, just go back to using reflection as @HendSayed25 suggested

ahmednasserzaza (approved):

iZyadNasser (approved):
Great job

HendSayed25 (approved):

PR #813, opened 2025-11-09
feat: add observeBalance function

Joseph-Sameh-0:


Ahmed7osny1 (approved):

Malak187 (approved):

nourelhodaahmed (approved):

wallet_data/src/commonMain/kotlin/net/thechance/mena/wallet/data/repository/balance/BalanceRepositoryImpl.kt
```diff
@@ -11,9 +18,24 @@ class BalanceRepositoryImpl(
...
+    }
+
+    override fun observeBalance(): Flow<Double> {
+        if (balanceFlow.value == 0.0) {
+            CoroutineScope(Dispatchers.IO).launch {
+                getBalance()
+            }
+        }
+        return balanceFlow
+    }
```
ZainabAlnajjar, line 34: sorry to interrupt another team PR but this case is shared for my team too and I have a little concern here
what happens when failed to fetch balance?
also, could you clarify what would happen if the user balance became 0.0 after another value? 

Imagine this case: 
User would get there balance and lets say it is 7.0, it was emitted in flow
then they transferred money or bought something with 7.0, the balance would be 0.0 in that case, but the flow won't be updated?

Joseph-Sameh-0: if the fetch failed there is an error will be thrown

if the balance becomes 0.0 nothing will change 
this check is only in first call if it 0.0 the function will try to update the balance
@ZainabAlnajjar

Malak187 (approved):

MuhammedEdrees (approved):

Malak187 (approved):

PR #826, opened 2025-11-10
fix cache of chat messages

Radwa85:


core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/repository/MessageRepositoryImpl.kt
```diff
@@ -76,27 +76,23 @@ class MessageRepositoryImpl(
...
             offset = (page * pageSize),
             limit = pageSize
-        )
-        val messages = cachedMessages.map { it.toDomain() }
-
-        if (messages.isEmpty()) {
-            return getFromRemote(chatId, page, pageSize)
-        }
-
-        val now = Clock.System.now().toString()
```
dev-bilal-azzam, line 86: why you removed this?
this way the `lastSyncTime` will never be stored, So we will never get last messages unless we request page 0 from remote.

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/repository/MessageRepositoryImpl.kt
```diff
@@ -76,27 +76,23 @@ class MessageRepositoryImpl(
...
+        ).map { it.toDomain() }
 
         val totalCachedItems = cachedMessageDao.getTotalMessagesCount(chatId.toString())
-        return PagedData(
-            data = messages,
+        val result = PagedData(
+            data = cachedMessages,
             totalItems = totalCachedItems,
-            isLastPage = false,
+            isLastPage = cachedMessages.size < pageSize
```
dev-bilal-azzam, line 85: I think the remote should tell if this is last page or not, because it might have pages that is not in  cache.

dev-bilal-azzam (changes requested):

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/repository/MessageRepositoryImpl.kt
```diff
@@ -76,27 +76,23 @@ class MessageRepositoryImpl(
             chatId = chatId.toString(),
             offset = (page * pageSize),
             limit = pageSize
-        )
```
dev-bilal-azzam, line 79:
```suggestion
)
        val messages = cachedMessages.map { it.toDomain() }

        if (messages.isEmpty()) {
            return getFromRemote(chatId, page, pageSize)
        }

        scope.launch(Dispatchers.IO) {
            val now = Clock.System.now().toString()
            val lastSyncTime = chatSyncTimeDao.getLastSyncTime(chatId.toString())
            if (lastSyncTime != null) {
                syncAfterLastUpdate(chatId)
            } else {
                chatSyncTimeDao.upsert(ChatSyncTime(chatId.toString(), now))
            }
        }

        val totalCachedItems = cachedMessageDao.getTotalMessagesCount(chatId.toString())
        return PagedData(
            data = messages,
            totalItems = totalCachedItems,
            isLastPage = false,
        )

dev-bilal-azzam: I think this will solve the bug, without affecting other aspects.

Al-Qassim:
please check the cache on the develop, if it works now or not. and make sure we still need this pr

Radwa85:
Yes i checked it , i think someone else solved  it so will close the pr

PR #833, opened 2025-11-10
fix reload product details on configuration change

haideraljawasim:
# What have been done :
- Fix reload product details on configuration change

### DIFF
--- a/dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/productDetails/ProductDetailsScreen.kt
+++ b/dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/productDetails/ProductDetailsScreen.kt
@@ -1,7 +1,6 @@
 package net.thechance.mena.dukan.presentation.screen.productDetails
 
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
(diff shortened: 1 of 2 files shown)


dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/productDetails/ProductDetailsScreen.kt
```diff
@@ -18,10 +17,6 @@ fun ProductDetailsScreen(
     val state by viewModel.state.collectAsStateWithLifecycle()
     val navController = LocalNavController.current
 
-    LaunchedEffect(Unit){
-        viewModel.refreshData()
-    }
-
```
YoussefmSaber, line 24: did you test the following, you add product to the cart, you go to the cart and delete the product from the cart, you click back and see the quantity is it 0 or it has the same quantity before removing it from the cart

haideraljawasim: I know that might stop updating the items on back from cart  , but what was here not the solution for the problem , also i see it normal to back and see the items still selected

mostafa-n3ma (approved):

AbdelrhmanAyman20470 (approved):

ahmedNaser7 (approved):

PR #836, opened 2025-11-10
Fix/dkn 125 navigate from dukan details small image to dukan map

Mohamed-Ashraf-2210:
Screenshot_20251110_202423_MENA

### DIFF
--- a/dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/dukanLocation/DukanLocationScreen.kt
+++ b/dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/dukanLocation/DukanLocationScreen.kt
@@ -31,6 +31,7 @@ import net.thechance.mena.dukan.presentation.viewModel.dukanLocation.DukanLocati
 import org.jetbrains.compose.resources.painterResource
 import org.jetbrains.compose.resources.stringResource
 import org.koin.compose.viewmodel.koinViewModel
+import org.maplibre.compose.camera.CameraPosition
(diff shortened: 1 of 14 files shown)


dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/dukanLocation/DukanLocationScreen.kt
```diff
@@ -0,0 +1,110 @@
...
+@Composable
+private fun DukanLocationContent(
+    state: DukanLocationUiState,
+    listener: DukanLocationInteractionListener
+) {
+    OnSystemBackPressed(listener::onBackClicked)
+    Scaffold(
+        topBar = { DukanLocationTopBar(onBackClick = listener::onBackClicked) }
+    ) {
+        val cameraState = rememberCameraState(state.cameraPosition)
```
YoussefmSaber: how about animate the camera so it feels more dynamic instead of showing it directly to the location
maybe like this:
```
val cameraState = rememberCameraState(firstPosition = CameraPosition())
LaunchedEffect(Unit) {
    cameraState.animateTo(
        finalPosition = cameraState.position.copy(
            target = state.cameraPosition,
            zoom = 16.0
        ),
        duration = 2.seconds
    )
}
```

YoussefmSaber (approved):
Good work, but take a look at my comment, and also animate the pointer to display after 1.5 seconds if you going to animate the camera

NadaMansour20 (approved):

ahmedNaser7 (approved):

NadaMansour20 (approved):

PR #837, opened 2025-11-10
refactor: improve error handling in DukanCartViewModel

NadaMansour20:
fix bug random error message appear when increment or decrement product quantity in cart screen

### DIFF
--- a/dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/viewModel/dukanCart/DukanCartViewModel.kt
+++ b/dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/viewModel/dukanCart/DukanCartViewModel.kt
@@ -221,9 +221,11 @@ class DukanCartViewModel(
     }
 
     private fun onErrorUpdateProductQuantity(throwable: Throwable) {
-        if(throwable is NoInternetException){
-            showSnackBar(message = Res.string.no_internet_connection)
+        val messageRes = when (throwable) {
+            is NoInternetException -> Res.string.no_internet_connection
+            else -> Res.string.something_went_wrong
         }
+        showSnackBar(message = messageRes)
     }
 
     private fun showSnackBar(message: StringResource, type: SnackBarType = SnackBarType.ERROR) {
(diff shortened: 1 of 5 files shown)


YoussefmSaber (approved):

Mohamed-Ashraf-2210 (approved):

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/viewModel/dukanCart/DukanCartViewModel.kt
```diff
@@ -221,11 +221,9 @@ class DukanCartViewModel(
...
 
     private fun onErrorUpdateProductQuantity(throwable: Throwable) {
-        val messageRes = when (throwable) {
-            is NoInternetException -> Res.string.no_internet_connection
-            else -> Res.string.something_went_wrong
+        if(throwable is NoInternetException){
+            showSnackBar(message = Res.string.no_internet_connection)
         }
-        showSnackBar(message = messageRes)
     }
```
haideraljawasim, line 229: hmmmm , so what the thrown exception was  , cause here we don't actually fix the problem we just hide the message

NadaMansour20: The thrown exception in this case is a CancellationException that happens when the previous coroutine is canceled when the user clicks multiple times quickly (because of the debounce)
Since the new request will still be sent after the debounce delay, we don’t need to show this error to the user.

ahmedNaser7 (approved):

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/viewModel/dukanCart/DukanCartViewModel.kt
mostafa-n3ma: I don't see any fix here we are handling only the no internet ex and ignoring else case !!

haideraljawasim (approved):

mostafa-n3ma (approved):

PR #846, opened 2025-11-11
fix/DKN-153-enable-adding-products-to-cart-from-details-screen-opened-from-search-results

mostafa-n3ma:
refactor product search response dto with dukan id 
so we can pass to the details viewmodel and add it latter to cart


AbdelrhmanAyman20470 (approved):

YoussefmSaber (approved):

haideraljawasim (approved):

ahmedNaser7:
i think the best solution, should be take dukanId from product details response not in navigation route , we  dont need dukan id in search screen

PR #855, opened 2025-11-11
fix: issue with the voice message not stopping on finish, and crash on direct second play

Al-Qassim:
new result

https://github.com/user-attachments/assets/d8612ea0-55f1-4177-ab61-cd82d20f7416

issue with the old code is that the `currentPositionSeconds` variable will never reach `duration`, if for example the `duration` was 2500, the `currentPositionSeconds` will reach a maximum of 2450, so the break condition will never pass

old code:
image

https://github.com/user-attachments/assets/9bc1c254-e4d8-4e49-9343-03bf107bd694

### DIFF
--- a/core_chat_domain/src/commonMain/kotlin/net/thechance/mena/core_chat/domain/model/SyncState.kt
+++ b/core_chat_domain/src/commonMain/kotlin/net/thechance/mena/core_chat/domain/model/SyncState.kt
@@ -5,7 +5,7 @@ import kotlin.uuid.ExperimentalUuidApi
 import kotlin.uuid.Uuid
 
 sealed class SyncState {
-    object Offline : SyncState()
+    data object Offline : SyncState()
     data class ChatsSummariesSynced(val chatSummaries: List<ChatSummary>) : SyncState()
 
     @OptIn(ExperimentalUuidApi::class)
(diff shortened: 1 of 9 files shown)


core_chat_domain/src/commonMain/kotlin/net/thechance/mena/core_chat/domain/model/SyncState.kt
```diff
@@ -5,7 +5,7 @@ import kotlin.uuid.ExperimentalUuidApi
 import kotlin.uuid.Uuid
 
 sealed class SyncState {
-    object Offline : SyncState()
+    data object Offline : SyncState()
```
ZainabAlnajjar, line 8: Why data object?

ZainabAlnajjar (approved):
Nice work, just please don't forget to clean up the not needed println from all files

core_chat_domain/src/commonMain/kotlin/net/thechance/mena/core_chat/domain/model/SyncState.kt
Al-Qassim: it's generally better, for the same reasons that we may prefer data class over plain class

raouf-kamel (approved):

dev-bilal-azzam (approved):

ZainabAlnajjar (approved):

PR #872, opened 2025-11-11
refactor: Use LazyVerticalGrid for product lists

ElsayedMagdy122:
Adaptive Products in Search 
image

Adaptive products in no image dukan
image

Adaptive products in small image dukan
image

### DIFF
--- a/dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/dukanDetails/components/noImageDukanDetails/NoImageDukanShelfWithProducts.kt
+++ b/dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/dukanDetails/components/noImageDukanDetails/NoImageDukanShelfWithProducts.kt
@@ -1,8 +1,13 @@
 package net.thechance.mena.dukan.presentation.screen.dukanDetails.components.noImageDukanDetails
 
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.lazy.grid.GridCells
+import androidx.compose.foundation.lazy.grid.LazyVerticalGrid
+import androidx.compose.foundation.lazy.grid.items
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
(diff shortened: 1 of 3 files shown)


dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/dukanDetails/components/smallImageDukanDetails/SmallImageDukanShelvesContent.kt
```diff
@@ -112,32 +116,43 @@ private fun ShelfProducts(
...
+                        key(product.id) {
+                            ProductItem(
+                                product = product,
+                                cartColor = cartColor,
+                                listener = listener
+                            )
+                        }
                     }
                 }
             }
```
YoussefmSaber, line 158: instead of using lazy raw with column with pair, just use lazy horizontal grid

YoussefmSaber (approved):

dukan_presentation/src/commonMain/kotlin/net/thechance/mena/dukan/presentation/screen/dukanDetails/components/noImageDukanDetails/NoImageDukanShelfWithProducts.kt
```diff
@@ -48,14 +53,22 @@ fun NoImageDukanShelfWithProducts(
...
-                listener = listener,
-                dukanColor = Color(dukanColor),
-                modifier = Modifier.padding(top = topPadding)
-            )
+        val spacing = Theme.spacing._8
+        val minCardWidth = 320.dp
+
+        LazyVerticalGrid(
+            modifier = Modifier.heightIn(min = 0.dp, max = 10_000.dp),
+            columns = GridCells.Adaptive(minSize = minCardWidth),
```
haideraljawasim, line 61: do we have  a nested  vertical scroll ? why you needed to have the `heightIn `modifier here  and what if the list execute the 10_000dp  , if not needed deleted other than this we need to search for a better solution

Ahmed-AlMancy: @elsayed magdy I think yeah there is a better solution I think it was made at wide image dukan screen

ElsayedMagdy122: Yes, having a nested vertical scroll here is tricky. The heightIn with 10_000dp was just a workaround to make the grid render fully inside a scrollable Column, but it’s not ideal. Removing it without a proper restructuring will break the layout, so a better solution would require rethinking the parent-child scroll structure, which is not straightforward and could be tricky to implement right now.

Mohamed-Ashraf-2210 (approved):

NadaMansour20 (approved):

PR #873, opened 2025-11-11
feature/FTH-148-implement-islamic-calender-and-integrate-it

BassantM:
---

### ScreenRec

https://github.com/user-attachments/assets/20ea22a8-e097-4e8d-a895-b28f5861bfdb


faith_domain/src/commonMain/kotlin/net/thechance/mena/faith/domain/usecase/GetNextPrayerTimeUseCase.kt
```diff
@@ -0,0 +1,40 @@
...
+        else {
+            val tomorrow = now.plusDays(1)
+            val tomorrowPrayers =
+                prayerTimeRepository.getPrayerTimes(date = tomorrow, address = address)
+            emit(tomorrowPrayers.firstOrNull())
+        }
+    }
+
+    @OptIn(ExperimentalTime::class)
+    fun Instant.plusDays(days: Int): Instant {
```
Hasa9nAli, line 36: i think this is wrong use of extension function inside use case , in the better way we have two scenario: 
1-  you can make this extension is private if you not used just inside this use case, if you don't use just inside this use case i not recommended put it inside use case 
2- make the `plusDays` is private function inside **GetNextPrayerTimeUseCase**

faith_domain/src/commonMain/kotlin/net/thechance/mena/faith/domain/usecase/GetNextPrayerTimeUseCase.kt
```diff
@@ -0,0 +1,40 @@
...
+        val nextPrayer = todayPrayers.firstOrNull { it.time > now }
+
+        if (nextPrayer != null) emit(nextPrayer)
+        else {
+            val tomorrow = now.plusDays(1)
+            val tomorrowPrayers =
+                prayerTimeRepository.getPrayerTimes(date = tomorrow, address = address)
+            emit(tomorrowPrayers.firstOrNull())
+        }
+    }
```
Hasa9nAli, line 33: i think to make it more cleaner we need to separate this use case to small readable pieces to make it better readable. 
For example : 
```
@OptIn(ExperimentalTime::class)
    fun getNextPrayer(address: Address): Flow<PrayerTime?> = flow {
        val now = Clock.System.now()
        val nextPrayer = getNextPrayerFromDay(now, address)
            ?: getNextPrayerFromDay(now.plusDays(1), address)

        emit(nextPrayer)
    }

    @OptIn(ExperimentalTime::class)
    private suspend fun getNextPrayerFromDay(date: Instant, address: Address): PrayerTime? {
        val prayers = prayerTimeRepository.getPrayerTimes(date, address)
        return prayers.firstOrNull { it.time > Clock.System.now() }
    }
```

in this way we make the nextPrayer in very clean and readable and easy to maintenance  in the future

PR #888, opened 2025-11-12
fix:empty state appears at first launch

gehadepada:
issueempty state appears at first launch cuz is loading not assign to true at onStart()

solution
handle base view model to ensure the loading state is set to true immediately when getUsers() is called, rather than waiting for the coroutine to be scheduled on the IO dispatcher.

another solution 
we can just make on start() not suspendable at base view model


admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/base/BaseViewModel.kt
```diff
@@ -49,9 +49,11 @@ abstract class BaseViewModel<STATE, EFFECT>(initialState: STATE) : ViewModel() {
                 onFinish?.invoke()
             }
         }
+        inScope.launch {
```
nourelhodaahmed, line 52: why you move it to another scope?

i think now the `callee` would start with `onStart` and won't wait for it to finish first

PR #899, opened 2025-11-12
fix: Remove unnecessary imePadding from CurrentPasswordContent

ShroukMohamed16:


AhmadKharfan (approved):

eslam-99 (approved):

identity_presentation/src/commonMain/kotlin/net/thechance/mena/identity/presentation/screen/changePassword/components/CurrentPasswordContent.kt
```diff
@@ -76,7 +76,6 @@ fun CurrentPasswordContent(
             modifier = Modifier
                 .fillMaxWidth()
                 .padding(bottom = Theme.spacing._12, top = Theme.spacing._24)
-                .imePadding()
         )
     }
```
ABDULLAHHG, line 80: It might be better to find another way to remove the padding,
since removing `imePadding` will place the button behind the keyboard when keyboard is open,
so that the user should dismiss the keyboard before being able to pressing the button

ABDULLAHHG: we may consider add `imePadding` to `NewPasswordContent`

ABDULLAHHG (approved):

PR #911, opened 2025-11-13
Fix/idn 171 buttons height

mohamedshemees:
this pr fixes buttons fixed height issues and refactor components preview


AhmadKharfan (approved):

ShroukMohamed16 (approved):

design_system/src/commonMain/kotlin/net/thechance/mena/designsystem/presentation/component/button/NegativeButton.kt
```diff
@@ -49,7 +54,7 @@ fun NegativeButton(
             Theme.colorScheme.primary.onPrimaryBody,
             Theme.colorScheme.primary.onPrimary
         ),
-        modifier = modifier.height(48.dp)
+        modifier = modifier
```
eslam-99, line 57: Removing the height will make each usage is hassle to calculate the needed vertical padding to match the design height, which is basically 48.dp, I suggest using `.defaultMinSize(minHeight = 48.dp)` to avoid fixed 48.dp on larger font sizes, also to prevent the hassle of calculations, what do you think?

mohamedshemees: What calculations ?
I think we just need padding then any usage will match the font size easily
We have a default padding as the design 
No need to send the padding until there is a specific padding which still the expected behaviour

eslam-99: I mean the padding inside the button itself.
If you noticed before, the button was thin, as it needed padding in every usage

eslam-99 (approved):

iBareq (approved):

PR #917, opened 2025-11-13
Fix/wlt 161 fix view and share transactions

hagarabobakr:


Malak187 (approved):

Joseph-Sameh-0 (approved):

nourelhodaahmed (approved):

Ahmed7osny1 (approved):

MennaSrour (approved):

wallet_presentation/src/commonMain/kotlin/net/thechance/mena/wallet/presentation/screen/statement_details/components/PdfViewer.kt
```diff
@@ -50,7 +47,7 @@ fun PdfViewer(
             horizontalAlignment = Alignment.CenterHorizontally,
             contentPadding = PaddingValues(top = 16.dp, bottom = 88.dp)
         ) {
-            itemsIndexed(pages) { index, page ->
+            itemsIndexed(pages!!) { _, page ->
```
MuhammedEdrees, line 50: Null assertions are not safe we can replace it with orEmpty() and since we don't need index we can use normal items function

```suggestion
            items(pages.orEmpty()) { page ->
```

PR #923, opened 2025-11-13
Fix/handle get balance error in home screen

ZainabAlnajjar:
Hide balance when any error happen when updating it


raouf-kamel (approved):

AdelTamer35 (approved):

Al-Qassim:
what was causing the problem?

Al-Qassim (approved):

ZainabAlnajjar:
> what was causing the problem?

The wallet module have done a refactor to return `null` in the flow when any error happens while fetching balance instead of returning `0.0`, so we needed to handle the new changes in our module properly

PR #932, opened 2025-11-14
Stop audio playback when chat screen is not visible

Mohamed-Elhanafy:
https://github.com/user-attachments/assets/7d4941bf-6a17-4d8f-a8d2-8814b4020966

### DIFF
--- a/core_chat_presentation/src/commonMain/kotlin/net/thechance/mena/core_chat/presentation/screen/chat/ChatScreen.kt
+++ b/core_chat_presentation/src/commonMain/kotlin/net/thechance/mena/core_chat/presentation/screen/chat/ChatScreen.kt
@@ -16,6 +16,7 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.lazy.LazyListState
 import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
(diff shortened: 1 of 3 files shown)


dev-bilal-azzam (approved):

core_chat_presentation/src/commonMain/kotlin/net/thechance/mena/core_chat/presentation/screen/chat/ChatScreen.kt
```diff
@@ -219,6 +224,30 @@ fun ChatScreenContent(
...
+        val observer = LifecycleEventObserver { _, event ->
+            when (event) {
+                Lifecycle.Event.ON_PAUSE -> {
+                    viewModel.onStopAudioPlayback()
+                }
+                Lifecycle.Event.ON_STOP -> {
+                    viewModel.onStopAudioPlayback()
+                }
+                else -> {}
+            }
```
ZainabAlnajjar: maybe we can just use `if` instead of `when` since we only have one function to either run or not, like this:

```
            if (event is Lifecycle.Event.ON_PAUSE  || event isLifecycle.Event.ON_STOP) {
                    viewModel.onStopAudioPlayback()
            }
```

ZainabAlnajjar (approved):

Al-Qassim (approved):

ZainabAlnajjar (approved):

PR #946, opened 2025-11-14
Fix/TRN-127-Handle no internet connection in reels screen

### LINKED ISSUE #936: Issue: Trending – Still loading when there is no internet connection.
Image

ahmeddsamii:
- Displays a "No Connection" message when a network error occurs while playing a reel.
- Provides a "Retry" button to allow users to reload the reel.
- Implements the necessary logic in the `UserReelViewModel` to manage network error states.
- Updates the `VideoPlayer` for both Android and iOS to detect and report network-related errors, such as `UnknownHostException` or `NSURLErrorDomain` with a no-connection error code.
- Refactors the error dialog for failed reel deletion to be based on `isReelDeleted == false` instead of a generic `error != null` check.

### DIFF
--- a/trends_presentation/src/androidMain/kotlin/net/thechance/mena/trends/presentation/video_player/VideoPlayer.android.kt
+++ b/trends_presentation/src/androidMain/kotlin/net/thechance/mena/trends/presentation/video_player/VideoPlayer.android.kt
@@ -131,20 +131,19 @@ actual fun VideoPlayer(
                     override fun onPlayerError(error: PlaybackException) {
                         val cause = error.cause
 
-                        if (cause is UnknownHostException ||
-                            error.errorCode == PlaybackException.ERROR_CODE_IO_NETWORK_CONNECTION_FAILED
-                        ) {
-                            onNetworkError()
-                            return
-                        }
+                        when {
+                            cause is UnknownHostException ||
+                                    error.errorCode == PlaybackException.ERROR_CODE_IO_NETWORK_CONNECTION_FAILED -> {
+                                onNetworkError()
+                            }
 
-                        if (cause is HttpDataSource.InvalidResponseCodeException &&
-                            cause.responseCode == HTTP_UNAUTHORIZED_STATUS_EXCEPTION
-                        ) {
-                            onRequestRefresh()
-                            return
+                            cause is HttpDataSource.InvalidResponseCodeException &&
+                                    cause.responseCode == HTTP_UNAUTHORIZED_STATUS_EXCEPTION -> {
+                                onRequestRefresh()
+                            }
+
+                            else -> super.onPlayerError(error)
                         }
-                        super.onPlayerError(error)
                     }
 
                     override fun onPlaybackStateChanged(state: Int) {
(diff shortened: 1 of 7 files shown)


trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/user_reel/UserReelViewModel.kt
```diff
@@ -128,7 +129,16 @@ internal class UserReelViewModel(
         )
     }
 
-    private fun onGetRefreshVideoUrl(refreshedUrl: String, reelId: String){
+    override fun onClickRetry(reelId: String) {
+        updateState { copy(currentReelId = reelId, error = null) }
+        getFeedReals()
```
mohamedd-hassan: typo here should be `Reels`

ahmeddsamii: This isn't my change, but okay i will modify it

mohamedd-hassan (commented):
im a little bit confused, why in the case of a playback error do we get the entire page again? shouldn't we just get the current reel again? because imagine you're on reel 5 and you lose connection and clicking retry suddenly requests the entire page again

second thing in the request refresh url function this is technically all we need to retry right? since a playback error means we fetched the page correctly just failed to play the current reel so shouldn't we just refresh the url?

trends_presentation/src/androidMain/kotlin/net/thechance/mena/trends/presentation/video_player/VideoPlayer.android.kt
```diff
@@ -126,12 +129,22 @@ actual fun VideoPlayer(
...
+                            return
+                        }
+
+                        if (cause is HttpDataSource.InvalidResponseCodeException &&
+                            cause.responseCode == HTTP_UNAUTHORIZED_STATUS_EXCEPTION
+                        ) {
+                            onRequestRefresh()
+                            return
+                        }
+                        super.onPlayerError(error)
```
ahmedmahgoub0: maybe when is better here

ahmedmahgoub0 (approved):
👏👏

trends_presentation/src/androidMain/kotlin/net/thechance/mena/trends/presentation/video_player/VideoPlayer.android.kt
ahmeddsamii: Yes, you are right

mohamedd-hassan (approved):

HendSayed25 (approved):

trends_presentation/src/iosMain/kotlin/net/thechance/mena/trends/presentation/video_player/VideoPlayer.ios.kt
```diff
@@ -140,10 +141,17 @@ actual fun VideoPlayer(
             if (item?.status == AVPlayerItemStatusFailed) {
                 val error = item.error
                 if (error != null) {
-                    if (error.code == NSURLErrorBadServerResponse ||
-                        error.domain == "NSURLErrorDomain"
-                    ) {
-                        onRequestRefresh()
+                    when {
+                        error.domain == "NSURLErrorDomain" &&
+                                error.code.toInt() == -1009 -> {
```
iZyadNasser: "-1009" is a magic number, put it in a constant with an expressive name that explains what the value represents

iZyadNasser (approved):

yusufnasserdev (approved):

iZyadNasser (approved):

ahmednasserzaza (approved):

PR #960, opened 2025-11-15
Fix/FTH 180 Stop Repeating Initial State In Surah

marwanqashwa:
ensures that the `onInitialAyahScrolled` method does not trigger its logic if either the Ayah sound is currently playing or the player UI is visible, preventing unwanted state updates during playback or when the player is open.

Behavior update:

* Prevented `onInitialAyahScrolled` from updating state when the player is visible, in addition to when Ayah sound is playing, by adding a check for `uiState.value.isPlayerVisible`.


fatmahgazy (approved):

Hasa9nAli (approved):
Nice eaaaa Marwan bug fighter

Abdulrahman-Ragab-01 (commented):
Great job, but convert the branch name to lowercase because the CI fails

BassantM (approved):

PR #977, opened 2025-11-16
fix: update addBookmark method

marwanqashwa:
refactors the way ayah bookmarks are added by changing the `addAyahBookmark` method to no longer return a bookmark object, both in the domain interface and repository implementation. Instead, the method now only performs the action of adding a bookmark

### DIFF
--- a/faith_data/src/commonMain/kotlin/net/thechance/mena/faith/data/remote/service/BookmarkApiService.kt
+++ b/faith_data/src/commonMain/kotlin/net/thechance/mena/faith/data/remote/service/BookmarkApiService.kt
@@ -14,7 +14,7 @@ import net.thechance.mena.faith.data.remote.model.bookmark.AyahBookmarkDto
 interface BookmarkApiService {
 
     @POST("faith/ayah/bookmarks")
-    suspend fun addBookmark(@Body request: AddBookmarkRequest): Response<AyahBookmarkDto>
+    suspend fun addBookmark(@Body request: AddBookmarkRequest): Response<Unit>
 
     @GET("faith/ayah/bookmarks")
     suspend fun getBookmarks(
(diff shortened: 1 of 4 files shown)


faith_data/src/commonMain/kotlin/net/thechance/mena/faith/data/remote/service/BookmarkApiService.kt
```diff
@@ -14,7 +14,7 @@ import net.thechance.mena.faith.data.remote.model.bookmark.AyahBookmarkDto
 interface BookmarkApiService {
 
     @POST("faith/ayah/bookmarks")
-    suspend fun addBookmark(@Body request: AddBookmarkRequest): Response<AyahBookmarkDto>
+    suspend fun addBookmark(@Body request: AddBookmarkRequest): Response<Unit>
```
fatmahgazy, line 17: Why are you returning `Response<Unit>`? I don’t think it adds any value here.

marwanqashwa: we need this returning type `import de.jensklingenberg.ktorfit.Response`
 to handle response code and so on by ktorfit

fatmahgazy: you mean you need it for status code right?

marwanqashwa: just take a look to this file `net.thechance.mena.faith.data.utils.executeApiSafely`

marwanqashwa: Provides:
 - Access to raw response, status code, headers, and status message.
 - Convenient methods to get deserialized body for successful responses.
 - Error body for unsuccessful responses.

BassantM (approved):

Hasa9nAli (approved):
keep going Marwan

Abdulrahman-Ragab-01 (approved):
Great job 👏🏻

PR #985, opened 2025-11-17
Refactor ChatSummary Message Model and Update Mappers

Mohamed-Elhanafy:
Replaced the nested `ChatSummary.Message` data class with the more comprehensive `Message` entity from the domain layer. This change propagates through the data, presentation, and domain modules.

Key changes:
- Removed the `Message` data class from `ChatSummary`.
- Updated mappers in `core_chat_data` (remote and local) to use the `Message` entity.
- Modified `HomeViewModelTest` to align with the new `Message` entity structure for creating test data.
- Adjusted the `ChatUiState` mapper in `core_chat_presentation` to handle the `Message` entity and its `MessageContent`.


core_chat_presentation/src/commonMain/kotlin/net/thechance/mena/core_chat/presentation/screen/home/mapper.kt
```diff
@@ -16,7 +17,7 @@ fun ChatSummary.toUi(): ChatUiState {
     val statusMessages = getStatusMessages(lastMessage, unReadMessagesCount)
     val lastMessage = lastMessage?.let {
         ChatUiState.MessageUiState(
-            text = it.content,
+            text = (it.content as? MessageContent.Text)?.text.orEmpty(),
```
ZainabAlnajjar: This is wrong, previously, the api was returning a text representing the content, but now it return the message and we handle the presentation
Please create a mapper to map the messages based on its type, if it is text take it, else add a localized string represent message type (Audio, Photo, Ayah, etc...) 

And we could later add type parameters, if we wanted to also display Ayah message content, cuz it needs a special font family

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/source/remote/mapper/chatSummaryMapper.kt
```diff
@@ -30,13 +34,18 @@ private fun List<ChatSummaryDto>.toListOfChatSummary(): List<ChatSummary> {
...
 fun ChatSummaryDto.toDomain(): ChatSummary {
-    val lastMessage = lastMessage?.let {
-        ChatSummary.Message(
-            content = lastMessage.content,
-            sendAt = Instant.parse(lastMessage.sentAt).toLocalDateTime(TimeZone.currentSystemDefault()),
-            isMine = lastMessage.isMine
+    val lastMessage = lastMessage?.let { lastMsg ->
+        Message(
+            id = Uuid.random(),
+            senderId = Uuid.random(),
```
ZainabAlnajjar: Why these are random uuid? We should also refactor the dto to use the correct message type and return correct values

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/source/local/database/cachedChatSummary/CachedChatSummaryDtoMapper.kt
```diff
@@ -15,7 +18,7 @@ fun ChatSummary.toCached(): CachedChatSummaryDto {
         id = this.id.toString(),
         name = this.name,
         imageUrl = this.imageUrl,
-        lastMessageContent = this.lastMessage?.content,
+        lastMessageContent = (this.lastMessage?.content as? MessageContent.Text)?.text,
```
ZainabAlnajjar, line 23: This is also wrong, we could cache the message id and get the message from the other table, I think we can have better strategy for caching now

ZainabAlnajjar (changes requested):

PR #993, opened 2025-11-17
Fix/IDN-199 Fix refresh token failure loop

AhmadKharfan:
### Fix refresh token failure loop

- Summary: Stop refresh retries on 401; emit blank token and navigate to login.
- Changes:
  - On refresh 401 (invalid/expired refresh token), clear tokens and emit to observers.
  - Emit blank token in `clearAuthTokens()` to trigger login routing.
- Reason: Refresh token can be invalid due to expiration, revocation, rotation, or stale persisted token.

### DIFF
--- a/identity_data/src/commonMain/kotlin/net/thechance/mena/identity/data/repository/AuthenticationRepositoryImpl.kt
+++ b/identity_data/src/commonMain/kotlin/net/thechance/mena/identity/data/repository/AuthenticationRepositoryImpl.kt
@@ -15,6 +15,7 @@ import net.thechance.mena.identity.data.mapper.toDomain
 import net.thechance.mena.identity.data.utils.postJson
 import net.thechance.mena.identity.data.utils.safeWrapper
 import net.thechance.mena.identity.domain.entity.PhoneNumber
+import net.thechance.mena.identity.domain.exception.UnAuthorizedException
 import net.thechance.mena.identity.domain.model.AuthenticationTokens
 import net.thechance.mena.identity.domain.repository.AuthenticationRepository
 import kotlin.concurrent.Volatile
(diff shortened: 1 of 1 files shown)


ShroukMohamed16 (approved):

identity_data/src/commonMain/kotlin/net/thechance/mena/identity/data/repository/AuthenticationRepositoryImpl.kt
```diff
@@ -50,14 +51,19 @@ class AuthenticationRepositoryImpl(
...
+            val response: AuthenticationResponse = safeWrapper {
+                client.postJson(
+                    RefreshRequestDto(settings.refreshToken),
+                    REFRESH_ENDPOINT
+                )
+            }
+            saveTokens(response.toDomain(), shouldEmit = !isTemporaryTokenMode)
+            settings.accessToken
+        } catch (e: UnAuthorizedException) {
+            saveTokens(createEmptyTokens(), shouldEmit = true)
```
nabilBouzineDev, line 64: Could you make the shouldEmit boolean in a separate function  because it 's no the responsibility of `saveTokens` ?

AhmadKharfan: I didn’t understand what you meant
saveTokens needs the shouldEmit boolean as a parameter so it can decide whether it should emit or not

mohamedshemees (approved):

identity_data/src/commonMain/kotlin/net/thechance/mena/identity/data/repository/AuthenticationRepositoryImpl.kt
nabilBouzineDev: I known but the name is kind of misleading, 

I thought saveTokens will just save the token locally. That's why, I think we should separate them into two functions or rename the function to reflect what it actually does.

PR #1014, opened 2025-11-18
Fix/IDN-207 Clear registration state after login to prevent incorrect redirect

AhmadKharfan:
#### Problem
- Users were sometimes redirected to upload profile image screen after login instead of home
- Registration draft data was not being cleared after login/registration completion

#### Solution
- Clear registration state (`lastRegistrationPhoneNumber` and draft data) after successful login
- Clear draft data after successful registration completion
- Ensures clean state management across login and registration flows

#### Changes
- **LoginUseCase**: Added registration state cleanup after login
- **AccountCreatedViewModel**: Added draft data cleanup after registration
- Updated unit tests for new behavior

### DIFF
--- a/identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/useCase/LoginUseCase.kt
+++ b/identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/useCase/LoginUseCase.kt
@@ -4,12 +4,10 @@ import net.thechance.mena.identity.domain.entity.PhoneNumber
 import net.thechance.mena.identity.domain.exception.InvalidMobileNumberException
 import net.thechance.mena.identity.domain.exception.InvalidPasswordException
 import net.thechance.mena.identity.domain.repository.AuthenticationRepository
-import net.thechance.mena.identity.domain.repository.RegistrationDraftRepository
 import net.thechance.mena.identity.domain.useCase.validation.mobileNumber.MobileNumberValidator
 
  class LoginUseCase(
     private val authenticationRepository: AuthenticationRepository,
-    private val registrationDraftRepository: RegistrationDraftRepository,
     private val mobileNumberValidator: MobileNumberValidator
 ){
      suspend fun login(countryCode: String, number: String, password: String) {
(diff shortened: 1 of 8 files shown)


identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/useCase/LoginUseCase.kt
```diff
@@ -4,20 +4,30 @@ import net.thechance.mena.identity.domain.entity.PhoneNumber
 import net.thechance.mena.identity.domain.exception.InvalidMobileNumberException
 import net.thechance.mena.identity.domain.exception.InvalidPasswordException
 import net.thechance.mena.identity.domain.repository.AuthenticationRepository
+import net.thechance.mena.identity.domain.repository.RegistrationDraftRepository
 import net.thechance.mena.identity.domain.useCase.validation.mobileNumber.MobileNumberValidator
 
  class LoginUseCase(
     private val authenticationRepository: AuthenticationRepository,
+    private val registrationDraftRepository: RegistrationDraftRepository,
```
eslam-99: Is it necessary to clear temp data during login as well?
I don't think it's the login's responsibility to clear another feature's temp data.

AhmadKharfan: Actually we already clear it during registration but I temporarily added the cleanup here as well to fix an issue in case someone registered using the old code where the data wasn’t cleared If they log in again with the new code the draft data will be removed properly
But this is only a temporary workaround to clear the old data and I’ll remove it next week once the current issues are resolved

AhmadKharfan: Honestly I’m not completely sure and I think I might remove it now instead of waiting until next week because there’s no time left

eslam-99: Cleaning up old versions is a kind of startup configuration i think, like db migrations which happens on new releases at first startup
I agree with deleting it as it not necessary now

AhmadKharfan: done

shahad-97 (approved):

eslam-99 (approved):

ABDULLAHHG (approved):

AimanYosofiUlt (approved):

PR #1028, opened 2025-11-19
Fix/idn 176 address type selection

ShroukMohamed16:


identity_presentation/src/commonMain/kotlin/net/thechance/mena/identity/presentation/screen/addresses/addEditLocation/AddEditLocationScreenViewModel.kt
```diff
@@ -102,7 +92,9 @@ class AddEditLocationScreenViewModel(
     override fun onClickMap() {
         sendNewEffect(
             createNavigateToMapEffect(
-                addressModel = null,
+                addressModel = createAddressModelFromCurrentState(
```
eslam-99: This should be null, to make pick location screen acts properly, after your change it's supposed to open the map with 16x zoom, not the default zoom value

ABDULLAHHG (approved):

AhmadKharfan (approved):

mohamedshemees (approved):

PR #1039, opened 2025-11-19
Fix TRN-133 Add Dark State Icons

HendSayed25:
- Add dark icons for empty state and no internet connection


MahmoodTarek (approved):

fatimafaisalghazi (approved):

yusufnasserdev (changes requested):
Kindly use the `AppThemeService` service for getting the theme and observing it, as it's the intended API for exposure for other modules.

I believe we should refrain from depending on repositories from other features.

PR #1054, opened 2025-11-20
add full screen option to the scaffold

AimanYosofiUlt:
Add fullScreen option to the Scaffold component, allowing screens to render without system bar padding. 
When enabled, the Scaffold expands to use the entire display area, providing a true full-screen experience.

```
Scaffold(
    fullScreen = state.showFullScreen,
    ...
```

### DIFF
--- a/design_system/src/commonMain/kotlin/net/thechance/mena/designsystem/presentation/component/scaffold/Scaffold.kt
+++ b/design_system/src/commonMain/kotlin/net/thechance/mena/designsystem/presentation/component/scaffold/Scaffold.kt
@@ -44,9 +44,9 @@ fun Scaffold(
         modifier = modifier
             .fillMaxSize()
             .applyIf(hasBlur) { blur(4.dp) }
-            .background(statusBarColor)
             .applyIf(!fullScreen) {
-                windowInsetsPadding(WindowInsets.statusBars)
+                background(statusBarColor)
+                    .windowInsetsPadding(WindowInsets.statusBars)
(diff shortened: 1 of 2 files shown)


AhmadKharfan (approved):

mohamedshemees (approved):

ShroukMohamed16 (approved):

design_system/src/commonMain/kotlin/net/thechance/mena/designsystem/presentation/component/scaffold/Scaffold.kt
```diff
@@ -41,13 +43,13 @@ fun Scaffold(
     Box(
         modifier = modifier
             .fillMaxSize()
-            .then(
-                if (hasBlur) Modifier.blur(4.dp) else Modifier
-            )
+            .applyIf(hasBlur) { blur(4.dp) }
             .background(statusBarColor)
```
mohamedshemees: i think this should be background color not statusBarColor , what do you thing ? , also where is it's usage ?  the statusBarColor

AimanYosofiUlt: There is a background color already!
But maybe the modifier needs to move inside the `applyIf `scope. I noticed it after your review 👍

ABDULLAHHG (approved):

iBareq (approved):

PR #1074, opened 2025-11-20
refactor: create generic dialog for Dukan status changes

Ahmed7osny1:
https://thechance-workspace.slack.com/archives/C09AX4J62LR/p1763679759721739


admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/component/DukanStatusChangeDialog.kt
```diff
@@ -0,0 +1,248 @@
...
+    cancelButtonText: String = stringResource(Res.string.cancel),
+    icon: Painter = painterResource(Res.drawable.ic_store_remove),
+    iconBackgroundColor: Color = Theme.colorScheme.background.bgError,
+    iconTint: Color = Theme.colorScheme.error,
+    isConfirmButtonEnabled: Boolean,
+    isConfirmButtonLoading: Boolean,
+    reasonPlaceholder: String? = null,
+    maxReasonLines: Int = 6,
+    modifier: Modifier = Modifier
+) {
```
gehadepada, line 52: it's better to put mandatory parameters first then optional + modifier should be first optional parameter

gehadepada: I think we don't need some parameters from these

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/dukan_details/component/DukanDetailsCard.kt
```diff
@@ -135,50 +135,22 @@ private fun DukanLocationMap(
...
+            )
         }
     }
-}
-
-@Composable
-private fun DukanLocation(
-    location: String,
-    modifier: Modifier = Modifier
-) {
```
nourelhodaahmed, line 164: why you removed this component?
we need it to view the location title

nourelhodaahmed (approved):

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/dukan_details/component/DukanDetailsCard.kt
Malak187: I think it is duplicated so he imported the other one in this file and used it, but the problem is that it left a large space between the name and the location so it needs adjustment

Image

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/dukan_details/DukanDetailsScreen.kt
```diff
@@ -108,9 +141,9 @@ private fun DukanDetailsScreenContent(
...
 private fun onDukanDetailsEffect(
     effect: DukanDetailEffect,
-    navController: NavController
+    onNavigateBack: () -> Unit
 ) {
     when (effect) {
-        DukanDetailEffect.NavigateBack -> navController.navigate(DukanManagement)
+        DukanDetailEffect.NavigateBack -> onNavigateBack()
     }
 }
```
Malak187, line 149: why did you make this change? it is fine to pass the nav controller here as we do in all screens

MuhammedEdrees (approved):

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/component/DukanStatusChangeDialog.kt
```diff
@@ -38,48 +36,81 @@ import org.jetbrains.compose.resources.painterResource
 import org.jetbrains.compose.resources.stringResource
 
 @Composable
-internal fun ScaffoldScope.RejectionDukanDialog(
+internal fun ScaffoldScope.DukanStatusChangeDialog(
```
MennaSrour, line 39: for this component, it looks around 90% similar to `AdminConfirmationDialog` so I suggest merging them into a single reusable dialog component and wrapping the reason TextField with a condition, using a parameter like:
`hasReasonField: Boolean = false`
This will reduce the redundant code and allow us to use one unified dialog for all admin panel screens

MennaSrour (approved):

Malak187 (approved):
Well done

PR #1106, opened 2025-11-22
Feature/ TRN -139 update Profile Placeholder 

MahmoodTarek:
# What I did in this PR:
- update Profile Placeholder
## Preview

| Mode | Before | After |
|------|--------|-------|
| **Light** |   Screenshot from 2025-11-22 12-30-35|  |
| **Dark** |Screenshot from 2025-11-22 12-30-10 | Screenshot from 2025-11-22 13-21-41 |

### DIFF
--- a/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/manage_my_trends/ManageTrendsScreen.kt
+++ b/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/manage_my_trends/ManageTrendsScreen.kt
@@ -302,8 +302,8 @@ private fun UserAvatar(
             model = profileImageUrl,
             contentDescription = stringResource(Res.string.profile_image_desc),
             error = tintedErrorPainter,
-            contentScale = ContentScale.Crop,
-            modifier = Modifier.clip(CircleShape)
+            modifier = modifier.size(100.dp).clip(CircleShape),
+            contentScale = ContentScale.Crop
         )
     }
 }
--- a/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/manage_my_trends/ManageTrendsScreen.kt
+++ b/trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/manage_my_trends/ManageTrendsScreen.kt
@@ -314,7 +314,7 @@ private fun EmptyProfilePicture(
     defaultPainter: Painter
 ) {
     val isDarkTheme = LocalDarkTheme.current
-    val backgroundColor = if (isDarkTheme) Color.Black else Color.White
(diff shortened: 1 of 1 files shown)


HendSayed25 (approved):

ahmedmahgoub0 (approved):

yusufnasserdev (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/manage_my_trends/ManageTrendsScreen.kt
```diff
@@ -288,13 +294,50 @@ private fun UserAvatar(profileImageUrl: String, modifier: Modifier = Modifier) {
...
+                shape = CircleShape,
+                ambientColor = iconTint.copy(alpha = 0.70f),
+                spotColor = iconTint.copy(alpha = 0.70f)
+            )
+            .clip(CircleShape)
+            .background(backgroundColor)
+            .padding(16.dp)
+
+    ) {
+        Icon(
```
ahmeddsamii: It's not right to the wrap the icon with box, use the modifier of the icon to make what you want

ahmeddsamii (changes requested):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/manage_my_trends/ManageTrendsScreen.kt
MahmoodTarek: I usually avoid wrapping Icon, but here using a Box ensures the avatar keeps a consistent size, padding, and shadow across Light/Dark themes without manually adjusting the Icon’s size
makes it more reliable and easier to maintain.

ahmeddsamii: It's not about readable or not, it's wrong to add unnecessary component because of the performance, and u can achieve the same ui with only icon

MahmoodTarek: I understand your point about performance and unnecessary components.
The reason I used a Box here is because applying padding, background, and shadow directly on the Icon causes it to expand unpredictably in Compose 
Using a Box ensures a fixed size container, keeps the Icon scaled properly, and avoids manual calculations for padding

MahmoodTarek: image

ahmeddsamii: U can give it size!

PR #1107, opened 2025-11-22
Fix TRN 137 Resolve incorrect video selection on back navigation

HendSayed25:
- Save the position of video that selected before play it to when navigate back return to the same position by smooth animation


trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/home/HomeScreen.kt
```diff
@@ -97,6 +104,22 @@ private fun HomeScreenContent(
...
+                    listener.saveScreenPosition(index, offset)
+                }
+        }
+
+        LaunchedEffect(reels.loadState.refresh) {
+            if (reels.loadState.refresh is LoadState.NotLoading) {
+                while (reels.itemCount <= state.savedIndex) {
+                    delay(10)
+                }
+                listState.animateScrollToItem(state.savedIndex, state.savedOffset)
```
iZyadNasser: I don't think it's right to use animation here, we don't want the user to see the transition, they only need to find themselves where they clicked when they go back

MahmoodTarek (approved):

trends_presentation/src/commonMain/kotlin/net/thechance/mena/trends/presentation/screen/home/HomeScreen.kt
HendSayed25: when remove it the scroll happen suddenly

PR #1108, opened 2025-11-22
Fix/IDN-238 Fix login after logout uses previous account

AhmadKharfan:
Switching accounts after logout reused the previous session due to Ktor Auth bearer token caching. Clear the cache on login, logout, and refresh; also clear local user on logout.

#### Changes
- Add `HttpClient.invalidateAuthTokens()` in common to clear bearer cache.
- Call `invalidateAuthTokens()` after `login`, `logout`, `refreshAccessToken`.
- Clear local user data on `logout`.


ShroukMohamed16 (approved):

mohamedshemees:
i think there is no need to call it after calling `refreshAccessToken`  because that's already what triggers the tokens refresh

ABDULLAHHG (approved):

eslam-99 (approved):

PR #1121, opened 2025-11-23
fix: enhance approve and reject animations in Dukan requests

Joseph-Sameh-0:
https://github.com/user-attachments/assets/b1fa1df8-3e4d-4537-91ae-293ab15d8cf5

### DIFF
--- a/admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/dukan_requests/DukanRequestsViewModel.kt
+++ b/admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/dukan_requests/DukanRequestsViewModel.kt
@@ -23,7 +23,6 @@ import net.thechance.mena.admin_panel.resources.status_updated_title
 import org.koin.android.annotation.KoinViewModel
 import org.koin.core.annotation.Provided
 import kotlin.uuid.ExperimentalUuidApi
-import kotlin.uuid.Uuid
 
 @OptIn(ExperimentalUuidApi::class)
 @KoinViewModel
@@ -80,7 +79,7 @@ class DukanRequestsViewModel(
                     message = currentState.rejectReason
                 )
             },
-            onSuccess = { onDukanApprovedSuccess(selectedDukanId) },
+            onSuccess = { onDukanApprovedSuccess() },
             onError = ::onError,
             dispatcher = dispatcher
         )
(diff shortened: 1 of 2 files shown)


MennaSrour (approved):

gehadepada (approved):

Farah315 (commented):
great work joseph 👏🏻👏🏻

Farah315 (approved):

nourelhodaahmed (approved):

EL-MANCY (approved):

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/dukan_requests/component/DukanRequestsTableContent.kt
```diff
@@ -96,12 +96,13 @@ private fun DukanListTable(
         state = listState,
         modifier = modifier
     ) {
-        itemsIndexed(items = dukan) { index, dukanItem ->
+        itemsIndexed(items = dukan, key = { index, item -> index to item.hashCode() }) { index, dukanItem ->
```
MuhammedEdrees: Why are we making the index of the item part of the key since it can change when an item is removed? Also can we simplify this by  using the item id instead of hashing the whole item?

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/screen/dukan_requests/DukanRequestsViewModel.kt
```diff
@@ -160,6 +161,14 @@ class DukanRequestsViewModel(
         }
     }
 
+    private fun removeDukanFromState(dukanId: Uuid) {
+        updateState {
+            it.copy(
+                dukans = it.dukans.filterNot { dukanItem -> dukanItem.id == dukanId }
+            )
+        }
+    }
```
MuhammedEdrees: The issue with this if there is multiple pages and an item is removed from current list it won't re-fetch the item so you'll have 7 items for example in the second page but if you go to second page there is an item that will not be fetched as it belongs to the first page and isn't visible there immediately after deleting. I think a better behaviour would be to re-fetch the current page after updating/deleting an item from the list.

Joseph-Sameh-0: I got it 
I will find a solution now

Malak187 (approved):

Malak187 (approved):

PR #1146, opened 2025-11-24
Feature/idn 247 support system appearance selection

mohamedshemees:
**this pr provides system theme appearance selection**
-- 
*   Added `DEFAULT` to the `AppTheme` enum to represent the system theme.
*   Updated the theme selection UI to include a `System` option with a descriptive hint.
*   Added new string resources for`System` and its hint in both English and Arabic.
*   Refactored `MenaTheme` to handle the `System` theme by checking `isSystemInDarkTheme`.
*   Simplified `AppThemeChanger` , removing manual system theme detection logic.
*   combine status and navigation bar appearance logic into a single `SetSystemBarsAppearance` 
notes : there is still some work to be done from `ios` side to make it reactive to system theme switching , currently it's working as on relaunching the app 

demo : 

https://github.com/user-attachments/assets/e8bf6f45-c070-40c8-991c-2249b3266cdd

**Edited:**

The issue of the IOS now is fixed by using this variable as a key  which will trigger recomposition only if the user chooses the system and the theme is in dark theme.

```kotlin
 val isFollowingSystemDarkMode = remember(appTheme, isSystemInDarkTheme) {
        appTheme == AppTheme.SYSTEM.name && isSystemInDarkTheme
    }
```

This way we will make the iOS side also trigger a recomposition when the `isSystemInDarkTheme()` updated  because now there's a running issue about this matter and you can refer to this issue [here](https://youtrack.jetbrains.com/issue/CMP-1986/isSystemInDarkTheme-should-dynamically-update-when-system-theme-is-changed)

https://github.com/user-attachments/assets/3633cbe4-0b18-4a09-98c9-bf26dd3092c4


ShroukMohamed16 (approved):

nabilBouzineDev (approved):
Well Done ✅

AimanYosofiUlt (approved):

poula22 (approved):

shahad-97 (approved):

iBareq (approved):

PR #1169, opened 2025-11-26
Fix/wlt 199 products column in dukan

MennaSrour:


admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/utils/PaginationTrigger.kt
```diff
@@ -26,4 +26,28 @@ fun <T> PaginationTrigger(
...
+            .distinctUntilChanged()
+            .collect { (lastVisibleIndex, totalCount) ->
+                lastVisibleIndex?.let {
+                    val lastIndex = if (totalCount > 0) totalCount - 1 else -1
+                    if (it >= lastIndex - buffer) {
+                        loadNextItems()
+                    }
+                }
+            }
+    }
```
MuhammedEdrees, line 30: Nice work. Can we use only this instead of having two `PaginationTrigger` ?

MuhammedEdrees (approved):
Well done 👏
As a small suggestion, we can replace padding on the whole content of the screen with padding on the components itself to avoid this clipping

Image

admin_panel/src/commonMain/kotlin/net/thechance/mena/admin_panel/presentation/utils/PaginationTrigger.kt
Joseph-Sameh-0: yes we can, 
can you do it? @MennaSrour

Joseph-Sameh-0 (approved):

wallet_data/src/commonMain/kotlin/net/thechance/mena/wallet/data/network_client/NetworkClient.kt
```diff
@@ -24,4 +24,6 @@ interface NetworkClient {
         urlString: String,
         requestBuilder: HttpRequestBuilder.() -> Unit = {}
     ): HttpResponse
+
+    fun clearCachedToken()
```
MuhammedEdrees, line 28: Currently this function is not used we need to call it when a user logs out

nourelhodaahmed (approved):

MuhammedEdrees (approved):

Ahmed7osny1 (approved):

PR #1171, opened 2025-11-26
Fix/cht 182 cache weather statement

yousef-osama11:
### **What was done in this PR:**
- add caching for weather in home screen

image


ZainabAlnajjar (approved):
Nice work, 
just one note, that you don't need to use `this` inside the mappers

Radwa85 (approved):

AdelTamer35 (approved):

PR #1183, opened 2025-11-27
Fix/cht 184 get current user in chat

yousef-osama11:
### **What was done in this PR:**
- depend on identity module to get user data instead of chat module


core_chat_presentation/src/commonMain/kotlin/net/thechance/mena/core_chat/presentation/screen/chat/ChatViewModel.kt
```diff
@@ -152,9 +152,17 @@ class ChatViewModel(
...
+                    if (user == null)
+                        onGetUserDataError()
+                    else
+                        onGetUserDataSuccess(user)
+
+                }
+            },
+            onError = { onGetUserDataError() }
         )
     }
```
ZainabAlnajjar, line 159: we can use `tryToCollect` here instead

yousef-osama11: try to collect doesn't support suspend functions

ZainabAlnajjar: In that case, I think we can move `onSuccess` logic inside `onGetUserDataSuccess`

ZainabAlnajjar (approved):

AdelTamer35 (approved):

Radwa85 (approved):

PR #1186, opened 2025-11-27
Fix/cht 185 reset http client after login

yousef-osama11:
What was done in this PR
- create a custom http client to reset client after logout and login

https://github.com/user-attachments/assets/db0c0de4-711f-4553-ab5f-3e19525900ce


core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/di/networkModule.kt
Al-Qassim, line 1: remove the old http client

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/source/remote/network/CustomHttpClient.kt
```diff
@@ -0,0 +1,36 @@
+package net.thechance.mena.core_chat.data.source.remote.network
+
+import io.ktor.client.HttpClient
+import io.ktor.client.engine.HttpClientEngineConfig
+import io.ktor.client.engine.HttpClientEngineFactory
+import kotlinx.serialization.json.Json
+import net.thechance.mena.identity.domain.service.AuthorizationService
+
+
+class CustomHttpClient(
```
Al-Qassim: don't name it custom http client, name it: HttpClientHolder

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/repository/UserRepositoryImpl.kt
```diff
@@ -1,17 +1,19 @@
...
 import net.thechance.mena.core_chat.data.source.remote.mapper.toDomain
+import net.thechance.mena.core_chat.data.source.remote.network.CustomHttpClient
 import net.thechance.mena.core_chat.data.source.remote.network.tryNetworkCall
 import net.thechance.mena.core_chat.domain.entity.User
 import net.thechance.mena.core_chat.domain.repository.UserRepository
 
 class UserRepositoryImpl(
-    private val client: HttpClient,
-    ): UserRepository {
+    private val customHttpClient: CustomHttpClient,
```
Al-Qassim: rename this: `private val clientHolder: HttpClientHolder`

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/repository/UserRepositoryImpl.kt
```diff
@@ -1,17 +1,19 @@
...
+) : UserRepository {
+
+    private val client = customHttpClient.getClient()
```
Al-Qassim: unfortunately this will not work, you should not store a copy of the client in the repository

Al-Qassim: when you reset the client inside the custom class, this copy will not be reset, it will have the same old client

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/repository/ChatRepositoryImpl.kt
```diff
@@ -203,6 +205,9 @@ class ChatRepositoryImpl(
             authRepository.observeTokenChange().collectLatest { token ->
                 if (token.isEmpty()) {
                     clearAllChatCache()
+                }else{
+                    customHttpClient.reset()
+                    client = customHttpClient.getClient()
```
Al-Qassim: here, you manually do a reset for the client copy of the chat repo, this is why you will not see issue with the home screen, but this is not a good practice.

you should always take the client from inside the custom class, and not store the client outside

Al-Qassim (changes requested):
please consider using the client inside the custom class as the only source of client, and do not save a copy anywhere else. these copies will not be updated

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/repository/MessageRepositoryImpl.kt
```diff
@@ -70,6 +70,7 @@ class MessageRepositoryImpl(
     private val messageSenderFactory: MessageSenderFactory,
     private val json: Json,
 ) : MessageRepository {
+    private val client = customHttpClient.getClient()
```
Al-Qassim: ok, I think you can solve this issue with the following, instead of 
`private val client = customHttpClient.getClient()`
write as 
```
private val client: HttpClient
    get() = customHttpClient.getClient()
```
this will not save a copy, and it will always get the client from the custom class

core_chat_data/src/commonMain/kotlin/net/thechance/mena/core_chat/data/source/remote/network/HttpClientHolder.kt
```diff
@@ -0,0 +1,89 @@
...
+    private val baseUrl: String,
+    private val authorizationService: AuthorizationService,
+    private val httpClientEngineFactory: HttpClientEngineFactory<HttpClientEngineConfig>,
+    private val json: Json
+) {
+    private var client: HttpClient = buildClient()
+
+    private fun buildClient(): HttpClient {
+        return createHttpClient()
+    }
```
Al-Qassim: no need for this function, delete it and use the other one directly

Al-Qassim (approved):

dev-bilal-azzam (approved):
Good Job Youssef, Can U please check the tests

Mohamed-Elhanafy (approved):

ZainabAlnajjar (approved):

AdelTamer35 (approved):
Good Job, We have heroes  👏👏

PR #1216, opened 2025-12-05
Fix/idn 101 camera not working and causing crash in edit profile screen

ABDULLAHHG:
### fix 

1 - fix camera not working and causing crash in edit profile screen  
2 - fix camera settings not open in ios 
3 - change the permission location behavioral in pick location screen to show snackbar for permission required when denied for the first time then navigate to enable permission screen on the second denied 

### refactor 
1 - implement camera permission native 
2 - remove moko permission
3 - refactor native permissions 
4 - clean presentation di 

<br>
Edit : `after video` has been updated with the implementation of before status and after status for native android
<br>

#### video that show the before and after for pick location screen 
<table>
  <tr>
    <td>
      <video src="https://github.com/user-attachments/assets/4788283e-090b-4ac6-a50f-7e011dba3d43" width="100%" controls></video>
      <p style="text-align:center; font-weight:bold;">Before</p>
    </td>
    <td>
      <video src="https://github.com/user-attachments/assets/b7b22c0b-ddc3-49bd-aa13-2d05aa0f0025" width="100%" controls></video>
      <p style="text-align:center; font-weight:bold;">After</p>
    </td>
  </tr>
</table>


identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/exception/Exception.kt
```diff
@@ -2,6 +2,7 @@ package net.thechance.mena.identity.domain.exception
 
 abstract class AuthenticationException(message: String) : Exception(message)
 abstract class LocationException(message: String) : Exception(message)
+abstract class Permission(message:String) : Exception(message)
```
AhmadKharfan: I think the permission cases shouldn’t be modeled as exceptions since they’re part of the normal user flow and not actual error conditions

identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/exception/Exception.kt
```diff
@@ -25,3 +26,8 @@ class InvalidRequestException() : AuthenticationException("Invalid request")
 class UnableToFindLocationException() : LocationException("Unable to find location")
 class CannotOpenSettingsException() : LocationException("Cannot open settings")
 class AddressNotFoundException() : LocationException("Address not found")
+
+class PermissionDeniedException() : Permission("Permission Denied")
+class PermissionDeniedPermanentlyException() : Permission("Permission Denied Permanently")
+class PermissionNotDeterminedException() : Permission("Permission Not Determined")
```
AhmadKharfan: > I think the permission cases shouldn’t be modeled as exceptions since they’re part of the normal user flow and not actual error conditions

Same here

identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/exception/Exception.kt
ABDULLAHHG: hmmmm i try to make it like that but in this case we will show a  error snackbar in a success function 
and that snackbar may say something like "location permission required" or may navigate the user to the enable location screen or settings 

so idk if we should handle these in success function not error i mean the snackbar 

i take a look on moko source code so i did it like that but we may change it to return permission state instead of throwing exception 

but still not sure about it so what do you think the best approach to go with ?

identity_presentation/src/iosMain/kotlin/net/thechance/mena/identity/presentation/util/permissions/CameraPermission.kt
```diff
@@ -0,0 +1,85 @@
...
+            AVAuthorizationStatusAuthorized -> PermissionState.GRANTED
+            AVAuthorizationStatusNotDetermined -> PermissionState.NOT_DETERMINED
+            AVAuthorizationStatusDenied -> PermissionState.DENIED
+            AVAuthorizationStatusRestricted -> PermissionState.DENIED_PERMANENTLY
+            else -> PermissionState.NOT_DETERMINED
+        }
+    }
+
+    private fun handleAuthorizationChange(
+        beforeStatus: AVAuthorizationStatus,
```
AhmadKharfan, line 56: Why do we need beforeStatus?

identity_presentation/src/iosMain/kotlin/net/thechance/mena/identity/presentation/util/permissions/CameraPermission.kt
```diff
@@ -0,0 +1,85 @@
...
+        afterStatus: AVAuthorizationStatus,
+        cont: CancellableContinuation<Unit>,
+        granted: Boolean
```
AhmadKharfan: Why rely on granted when afterStatus already tells the truth? The granted boolean is basically redundant because if granted == true the afterStatus will always be Authorized

identity_presentation/src/iosMain/kotlin/net/thechance/mena/identity/presentation/util/permissions/CameraPermission.kt
```diff
@@ -0,0 +1,85 @@
...
+                cont.resumeWithException(PermissionDeniedPermanentlyException())
+            }
+
+            afterStatus == AVAuthorizationStatusNotDetermined -> {
+                cont.resumeWithException(PermissionNotDeterminedException())
+            }
+
+            else -> {
+                cont.resumeWithException(PermissionDeniedException())
+            }
```
AhmadKharfan: The conditions result in a lot of duplicated branches
I think this enough

```suggestion
    when (afterStatus) {
        AVAuthorizationStatusAuthorized -> cont.resume(Unit)

        AVAuthorizationStatusDenied ->
            cont.resumeWithException(PermissionDeniedException())

        AVAuthorizationStatusRestricted ->
            cont.resumeWithException(PermissionDeniedPermanentlyException())

        AVAuthorizationStatusNotDetermined ->
            cont.resumeWithException(PermissionNotDeterminedException())

        else ->
            cont.resumeWithException(PermissionDeniedException())
    }

identity_presentation/src/iosMain/kotlin/net/thechance/mena/identity/presentation/util/permissions/CameraPermission.kt
ABDULLAHHG: ios dose not have the same mechanizem as the android 

ios have only one time request so i should show snackbar if the user deny on the first time 
and navigate to settings on the second time 

that why we use here before and after 

i am thinking about make the native android the same so when the user deny on request will only show the snackbar
the current implementation in android if the user deny the second will navigate directly to the settings

identity_domain/src/commonMain/kotlin/net/thechance/mena/identity/domain/exception/Exception.kt
AhmadKharfan: I see the cleanest approach is to avoid using exceptions for permission handling  permission results are part of the normal user flow not unexpected errors.
Instead of throwing exceptions the permission request function should return a permission state then the UI layer can decide what to do with each state show a snackbar, navigate to settings, request again, etc...

identity_presentation/src/iosMain/kotlin/net/thechance/mena/identity/presentation/util/permissions/CameraPermission.kt
AhmadKharfan: Yeah got you it would be good to make Android follow the same behavior so both platforms stay consistent

AhmadKharfan (approved):

haideraljawasim (approved):

MuhammedEdrees (approved):

